PR #4, opened 2017-04-18
Used WordPress function edit_post_link() instead of astra_edit_post_link()

maheshwaghmare:
- Removed function `astra_edit_post_link()` and used `edit_post_link()`.
- To disable the edit link on page builder. We can use the filter `edit_post_link`.


Nikschavan:
Closing this as this function is actually useful and there is no need to remove it.

PR #5, opened 2017-04-19
Used ast_get_option() instead of Ast_Theme_Options::get_options()

maheshwaghmare:
Function `get_options()` from class `Ast_Theme_Options` works exact same as `ast_get_option()` function.
So, Using `ast_get_option()` instead of `Ast_Theme_Options::get_options()`.


Nikschavan:
@maheshwaghmare the Class `Ast_Theme_Options ` is still used in the theme and if we are removing the usage of it, I expect the class also to be removed from the theme.

PR #10, opened 2017-04-19
Version 1.0.1.1

Nikschavan:


inc/core/class-ast-admin-helper.php
```diff
@@ -0,0 +1,87 @@
...
+		 * Returns an option from the database for
+		 * the admin settings page.
+		 *
+		 * @param  string  $key     The option key.
+		 * @param  boolean $network Whether to allow the network admin setting to be overridden on subsites.
+		 * @return string           Return the option value
+		 */
+		static public function get_admin_settings_option( $key, $network = false ) {
+			// Get the site-wide option if we're in the network admin.
+			if ( is_network_admin() ) {
```
Nikschavan: double if conditions can be removed here, Only this condition should work

```
if ( $network && is_multisite() ) {
	$value = get_site_option( $key );
} else {
	$value = get_option( $key );
}
```

The current condition is redundantly checking if we are on multisite.

inc/core/class-ast-admin-helper.php
```diff
@@ -0,0 +1,87 @@
...
+		 * Updates an option from the admin settings page.
+		 *
+		 * @param string $key 		The option key.
+		 * @param mixed  $value 	The value to update.
+		 * @param bool   $network 	Whether to allow the network admin setting to be overridden on subsites.
+		 * @return mixed
+		 */
+		static public function update_admin_settings_option( $key, $value, $network = false ) {
+			// Update the site-wide option since we're in the network admin.
+			if ( is_network_admin() || $network ) {
```
Nikschavan: double if conditions can be removed here, Only this condition should work

```
if ( $network && is_multisite() ) {
	$value = get_site_option( $key );
} else {
	$value = get_option( $key );
}
```

The current condition is redundantly checking if we are on multisite.

inc/core/class-ast-admin-helper.php
```diff
@@ -0,0 +1,87 @@
...
+		 * Returns an option from the database for
+		 * the admin settings page.
+		 *
+		 * @param string $key The option key.
+		 * @param bool   $network Whether to allow the network admin setting to be overridden on subsites.
+		 * @return mixed
+		 */
+		static public function delete_admin_settings_option( $key, $network = false ) {
+			// Get the site-wide option if we're in the network admin.
+			if ( is_network_admin() ) {
```
Nikschavan: double if conditions can be removed here, Only this condition should work

```
if ( $network && is_multisite() ) {
	$value = get_site_option( $key );
} else {
	$value = get_option( $key );
}
```

The current condition is redundantly checking if we are on multisite.

Nikschavan:
Just a few fixes @maheshwaghmare

assets/css/unminified/style-rtl.css
```diff
@@ -2349,6 +2349,8 @@ body {
 .main-header-menu {
   /*
 	> .ast-left-align-sub-menu {
```
Nikschavan: @maheshwaghmare Can you remove the commented code here?

assets/css/unminified/style.css
```diff
@@ -2349,6 +2349,8 @@ body {
 .main-header-menu {
   /*
 	> .ast-left-align-sub-menu {
+
```
Nikschavan: Some more commented code that can be removed.

PR #21, opened 2017-04-25
Chnaged default values and strings

rushijagani:
Default theme strings and values updated.


inc/core/class-ast-theme-options.php
```diff
@@ -131,7 +131,7 @@ public static function defaults() {
 				'disable-primary-nav'				=> false,
 				'header-layouts'					=> 'header-main-layout-1',
 				'header-main-rt-section'			=> 'none',
-				'header-main-rt-section-html'		=> __( 'HTML is allowed here' , 'astra' ),
+				'header-main-rt-section-html'		=> __( '<button>Contact Us</button>' , 'astra' ),
```
Nikschavan: Here, the HTML tags can be kept outside of the translatable text.

rushijagani: Html tag removed from the translatable text.

PR #27, opened 2017-05-03
Removed unused variables when a class is initialised

maheshwaghmare:


changelog.txt
```diff
@@ -1,4 +1,5 @@
 v1.0.1.2
+# Removed: Initialized objects of /compatibility/*.php classes.
```
Nikschavan: @maheshwaghmare update the changelog. `Improvement: Removed unused variables when a class is initialized`

maheshwaghmare: Updated `Initialized objects of /compatibility/*.php classes.` with `Improvement: Removed unused variables when a class is initialized`

PR #32, opened 2017-05-05
Beaver themer support

dineshchouhan:
Page builder implemented for BB and Beaver Themer.


sass/site/site-compatible/bb-plugin.scss
```diff
@@ -116,3 +116,17 @@
     clear: both;
 }
 
+a.fl-button,
+a.fl-button:visited,
```
Nikschavan: We should try to avoid using element tags in the CSS/SCSS.

Also, should the compiled version of this scss be in this PR?

PR #72, opened 2017-05-26
Improved sanitisation and updated theme prefix on places where it was missed

dineshchouhan:


inc/blog/blog.php
```diff
@@ -168,6 +168,7 @@ function astra_get_audios_from_post( $post_id ) {
 		// check what is the first embed containg video tag, youtube or vimeo.
 		foreach ( $embeds as $embed ) {
 			if ( strpos( $embed, 'audio' ) ) {
+				// Avoiding escaping intentionally for $embed.
 				return '<span class="ast-post-audio-wrapper">' . $embed . '</span>';
```
Nikschavan: @dineshchouhan `wp_kses()` can be used here allowing only required tags to be passed, instead of completely eliminating escaping.

PR #74, opened 2017-05-29
Merge V1.0.5.1

dineshchouhan:
v1.0.5.1
Fixed: Use `&hellip;` instead of `...`
Fixed: Use sanitize_html_class to sanitize HTML class
Fixed: Remove comment-list from add_theme_support( 'html5' )
Fixed: Escaping implemented for Dynamic CSS
Search result page escaping issue.

### DIFF
--- a/inc/class-astra-dynamic-css.php
+++ b/inc/class-astra-dynamic-css.php
@@ -139,17 +139,17 @@ if ( ! class_exists( 'Astra_Dynamic_CSS' ) ) {
 					'font-size' => astra_get_font_css_value( (int) $body_font_size['desktop'] * 6.25, '%' ),
 				),
 				'a, .page-title' => array(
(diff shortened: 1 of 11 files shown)


inc/class-astra-dynamic-css.php
```diff
@@ -296,28 +296,28 @@ static public function return_output() {
 				// Widget Title.
 				'.widget-title' => array(
 					'font-size' => astra_get_font_css_value( (int) $body_font_size['desktop'] * 1.428571429 ),
```
Nikschavan: does output of `astra_get_font_css_value ()` needs to be escaped?

dineshchouhan: No need to escape output because output sent by this function is already escaped.

PR #77, opened 2017-05-30
Used filter 'after_setup_theme' to set default theme options instead of action 'init'.

maheshwaghmare:
Removed unwanted function which was set the default customizer default options.


Nikschavan:
@maheshwaghmare Can you fix the merge conflict?

PR #98, opened 2017-06-07
Fixed: RTL support for minified css files.

maheshwaghmare:


changelog.txt
```diff
@@ -1,3 +1,6 @@
+{DEV}
```
Nikschavan: Instead of `{DEV}` add version number of milestone. Eg -   `v1.0.3 (unreleased)`

maheshwaghmare: Updated `{DEV} with `v1.0.5.3 (unreleased)` which is our upcomming release.

PR #99, opened 2017-06-07
Version 1.0.6

Nikschavan:
- Enhancement: Combined CSS files 'blog-layout-1' and 'astra-font' files in style.css.
- Enhancement: Combined JS files 'flexibility', 'navigation' and 'skip-link-focus-fix' files in style.js
- Enhancement: Used astra_get_option() to set default customizer options instead of representative Astra_Theme_Options::defaults().
- Fixed: RTL support for minified css files.
- New: Typography option for Archive Summary Box.
- Enhancement: Updated default heading font-size support for small devices.


inc/customizer/class-astra-customizer-sanitizes.php
```diff
@@ -80,6 +80,21 @@ static public function sanitize_number_n_blank( $val ) {
...
+		 * @return number        Return number.
+		 */
+		static public function sanitize_spacing( $val ) {
+
+			foreach ( $val as $key => $value ) {
+				$val[ $key ] = is_numeric( $val[ $key ] ) ? $val[ $key ] : '';
+			}
+
+			return $val;
+		}
```
Nikschavan: Function name `sanitize_spacing` is actually sanitizing number, also This does not seem to be used anywhere.

inc/customizer/class-astra-customizer-sanitizes.php
```diff
@@ -80,6 +80,21 @@ static public function sanitize_number_n_blank( $val ) {
 		}
 
 		/**
+		 * Sanitize Spacing
+		 *
+		 * @param  number $val Customizer setting input number.
+		 * @return number        Return number.
+		 */
+		static public function sanitize_spacing( $val ) {
```
Nikschavan: Function `sanitize_spacing` sanitizes number? also it seems this function is not used anywhere in the code?

dineshchouhan: @Nikschavan, Needs to move Spacing Param itself in Addon, when we move this into Addon in that time this sanitization function also moves into Addon.

PR #130, opened 2017-07-19
Container layout defaults value updated and Implemented backward compatibilty.

dineshchouhan:


inc/core/class-astra-theme-update.php
```diff
@@ -268,7 +273,33 @@ static public function v_1_0_8() {
...
+				'site-content-layout'         => 'plain-container',
+				'single-page-content-layout'  => 'plain-container',
+				'single-post-content-layout'  => 'content-boxed-container',
+				'archive-post-content-layout' => 'content-boxed-container',
+			);
+
+			$astra_options = get_option( ASTRA_THEME_SETTINGS, array() );
+
+			foreach ( $options as $key => $value ) {
+				if ( ! array_key_exists( $key, $astra_options ) ) {
```
Nikschavan: instead of `array_key_exists ` prefer `isset()` as this can be bit less expensive.

PR #142, opened 2017-07-31
Page builder compatibility

dineshchouhan:
Page Builders compatibility implemented.


inc/theme-update/class-astra-theme-update.php
```diff
@@ -313,9 +313,10 @@ static public function v_1_0_12() {
 		}
 
 		/**
-		 * Update options of older version than 1.0.13.
+		 * Update options of older version than 1.0.7.
 		 *
-		 * @since 1.0.13
+		 * @since 1.0.7
```
Nikschavan: Since should be 1.0.13

PR #226, opened 2017-09-19
Error message alignment fix

Nikschavan:


inc/assets/css/astra-admin-menu-settings.css
```diff
@@ -172,6 +172,7 @@
  */
 #message {
     display: block;
-    float: left;
-     width: 100%;
+    margin-left: 0;
+    /*float: left;
+    width: 100%;*/
```
Nikschavan: @sandeshjangam We should not ship commented code, let's remove this completely from the CSS file

PR #229, opened 2017-09-19
Astra fonts filter

rushijagani:


inc/class-astra-dynamic-css.php
```diff
@@ -598,6 +598,21 @@ static public function return_output() {
 			/* Parse CSS from array()*/
 			$parse_css .= astra_parse_css( $site_width, '769' );
 
+			/**
+			 * Astra Fonts
+			 */
+			if ( apply_filters( 'astra_fonts_support', true ) ) {
```
Nikschavan: Let's name this filter - `astra_enable_default_fonts` as `astra_font_support` sounds more like general purpose font support and disabling it should disable all the fonts.

rushijagani: astra_enable_default_fonts updated

PR #231, opened 2017-09-19
oEmbed responsive support

rushijagani:


inc/class-astra-after-setup-theme.php
```diff
@@ -143,6 +143,23 @@ function setup_theme() {
 			$dir_name    = ( SCRIPT_DEBUG ) ? 'unminified' : 'minified';
 			$file_prefix = ( SCRIPT_DEBUG ) ? '' : '.min';
 			add_editor_style( 'assets/css/' . $dir_name . '/editor-style' . $file_prefix . '.css' );
+
+			if ( apply_filters( 'astra_responsive_oembed', true ) ) {
```
Nikschavan: Let's rename this filter to `astra_fullwidth_oembed` as this is more relevant to what this does.

rushijagani: astra_fullwidth_oembed updatred

PR #248, opened 2017-10-11
Edge browser responsive menu support added

rushijagani:
v1.0.23  
**Fix:** Edge browser responsive navigation menu not working.


Nikschavan:
@rushijagani Have you tested this fix on actual devices as well for both - IE and EDGE?

PR #249, opened 2017-10-11
White label filter for welcome page details

rushijagani:
v1.0.23
Improvement: Added filter to support white-label for the theme name on theme's welcome page.

### DIFF
--- a/inc/core/view-general.php
+++ b/inc/core/view-general.php
@@ -9,6 +9,7 @@
  * @since       Astra 1.0
  */
 
+$astra_theme_name = apply_filters( 'astra_theme_name', __( 'Astra' , 'astra' ) );
 ?>
 
 <div class="wrap ast-clear">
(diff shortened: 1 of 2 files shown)


inc/core/view-general.php
```diff
@@ -19,10 +19,20 @@
...
+								?>
+							</span>
+						</h2>
 						<div class="inside">
 							<p>
-								<?php esc_html_e( 'Astra is a very lightweight and beautiful theme made to work with Page Builders.', 'astra' ); ?>
+								<?php
+								/* translators: 1: Astra Theme name */
+								printf( esc_html__( '%1$s is a very lightweight and beautiful theme made to work with Page Builders.', 'astra' ), apply_filters( 'astra_theme_name', __( 'Astra' , 'astra' ) ) );
+								?>
```
Nikschavan: Get the filtered theme name in a variable and then use that variable on multiple places instead of using same filter on multiple places

rushijagani: Yes, Right. Just added a variable for Astra Theme name filter

PR #313, opened 2017-12-15
Stock Availability style updated

dineshchouhan:


inc/compatibility/woocommerce/woocommerce-common-functions.php
```diff
@@ -128,11 +128,14 @@ function astra_woo_product_in_stock( $markup, $product ) {
 			if ( ! empty( $availability ) && $stock_quantity ) {
 				ob_start();
 				?>
-				<p class="stock in-stock">
-					<?php
-					/* translators: 1: in stock string */
-					printf( __( '<span class="ast-stock-avail">Availability:</span> %s', 'astra' ), $availability );
-					?>
+				<p class="ast-stock-detail">
+					<span class="ast-stock-avail">Availability:</span>
```
Nikschavan: This string needs to translated

inc/compatibility/woocommerce/woocommerce-common-functions.php
```diff
@@ -128,11 +128,14 @@ function astra_woo_product_in_stock( $markup, $product ) {
...
+					<span class="stock in-stock">
+						<?php
+						/* translators: 1: in stock string */
+						printf( __( '%s', 'astra' ), $availability );
```
Nikschavan: No need to translate the dynamic string.

PR #364, opened 2017-12-28
Astra filters deprecated

rushijagani:
- Fixed: Deprecated filter fatal error in WordPress 4.5 & below.

### DIFF
--- a/inc/core/deprecated/deprecated-filters.php
+++ b/inc/core/deprecated/deprecated-filters.php
@@ -25,7 +25,7 @@ add_filter( 'astra_color_palettes', 'deprecated_astra_color_palette', 10, 1 );
  */
 function deprecated_astra_color_palette( $color_palette ) {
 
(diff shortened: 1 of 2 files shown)


inc/core/deprecated/deprecated-filters.php
```diff
@@ -25,7 +25,7 @@
  */
 function deprecated_astra_color_palette( $color_palette ) {
 
-	$color_palette = apply_filters_deprecated( 'astra_color_palletes', array( $color_palette ), '1.0.22', 'astra_color_palettes', '' );
+	$color_palette = apply_astra_filters_deprecated( 'astra_color_palletes', array( $color_palette ), '1.0.22', 'astra_color_palettes', '' );
```
Nikschavan: Name the function `astra_apply_filters_deprecated` since our prefix for all the functions is `astra`

PR #417, opened 2018-01-29
LifterLMS basic compatibility Implemented.

dineshchouhan:
LifterLMS basic compatibility Implemented.


inc/compatibility/lifterlms/class-astra-lifterlms.php
```diff
@@ -0,0 +1,572 @@
...
+				if ( has_filter( 'llms_review_custom_styles' ) ) {
+					$styles = apply_filters( 'llms_review_custom_styles', $styles );
+				}
+
+				foreach ( $posts_array as $post ) {
+					echo $styles['custom-css'];
+
+					?>
+					<div class="llms_review" style="background-color:<?php echo $styles['background-color']; ?>;">
+						<h5 style="color:<?php echo $styles['title-color']; ?>;"><strong><?php echo get_the_title( $post->ID ); ?></strong></h5>
```
Nikschavan: We should escape the generated CSS values.

PR #470, opened 2018-03-01
Radio tabs param implemented.

dineshchouhan:


Nikschavan:
Closing since at this moment the customizer architecture is changed quite a lot with [Astra_Customizer](https://github.com/brainstormforce/astra/blob/master/inc/customizer/class-astra-customizer.php)

PR #476, opened 2018-03-12
remove post nav on LifterLMS quizzes

thomasplevy:
Remove the hooked `astra_single_post_navigation_markup()` for LifterLMS quizzes.

It's pretty clear that this kind of navigation is intended for blog posts (and maybe pages) but causes a poor user experience when displayed on quizzes.


Nikschavan:
Thank you very much for the PR @thomasplevy 👍

PR #478, opened 2018-03-13
Fixes issues related to video embeds due to ast-oembed-container class

Nikschavan:
- Fixed black bars when the video is made full width using the `ast-oembed-container` class
- Allow the `oembed` container class to pass through the filter `oembed_result` so that videos added through `wp_oembed_get()` will also get the wrapper.
- Introduces filter `astra_allowed_fullwidth_oembed_providers ` of providers to be made full width.

### DIFF
--- a/inc/class-astra-after-setup-theme.php
+++ b/inc/class-astra-after-setup-theme.php
@@ -168,7 +168,7 @@ if ( ! class_exists( 'Astra_After_Setup_Theme' ) ) {
 			$add_astra_oembed_wrapper = apply_filters( 'astra_responsive_oembed_wrapper_enable', true );
 
 			$allowed_providers = apply_filters(
-				'astra_allowed_fullwidth_providers', array(
+				'astra_allowed_fullwidth_oembed_providers', array(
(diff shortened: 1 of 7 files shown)


inc/class-astra-after-setup-theme.php
```diff
@@ -157,19 +158,31 @@ function setup_theme() {
...
+		 *
 		 * @return string       Updated embed markup.
 		 */
 		function responsive_oembed_wrapper( $html, $url, $attr ) {
 
 			$add_astra_oembed_wrapper = apply_filters( 'astra_responsive_oembed_wrapper_enable', true );
 
-			if ( false !== strpos( $url, 'vimeo.com' ) || false !== strpos( $url, 'youtube.com' ) || false !== strpos( $url, 'youtu.be' ) ) {
+			$allowed_providers = apply_filters(
+				'astra_allowed_fullwidth_providers', array(
```
Nikschavan: We should rename the filter to `astra_allowed_fullwidth_oembed_providers `

PR #496, opened 2018-03-23
Elementor pro 2.0 Support

Nikschavan:


inc/compatibility/class-astra-beaver-themer.php
```diff
@@ -186,6 +186,16 @@ function theme_header_footer_render() {
...
 						);
 					}
+
+					// Override! Footer Widgets.
+					$footer_widgets = get_post_meta( $template_id, 'footer-adv-display', true );
+					if ( 'disabled' === $footer_widgets ) {
+						add_filter(
+							'astra_advanced_footer_disable', function() {
+								return true;
+							}
```
Nikschavan: Instead of using anonymous function can we use `__return_true` here? anonymous functions are not supported by older versions of PHP

Nikschavan: We have many other anonymous functions for BB/Elementor compatibility, We can skip this as fixing this will not help with PHP support.

PR #477, opened 2018-03-12
add LifterLMS course builder layout option compatibility

thomasplevy:
This adds Astra layout post options to LifterLMS quizzes via the [course builder theme compatibility API](https://lifterlms.com/docs/quiz-theme-compatibility-developers/).

This allows users to select alternate layout options on LifterLMS quizzes via the builder.

course_builder_ _llms-72_dev_ _wordpress

Currently we only have support for a single field so possibly we could add support for the sidebar fields (as well as many of the other layout/post options available in Astra. However, at this moment I wanted to make this available to you if you were interested in using it.

<3

PR #546, opened 2018-04-30
add astra theme fields to LifterLMS builder settings

thomasplevy:
Per my note on #477 and your feedback we rebuilt our custom fields api to ensure themes could add a lot more custom fields to the builder.

I used Astra as one of the themes I tested against during development of a better set of functions and settings to allow more theme settings to be added to the builder. After getting everything working and debugging a bit I wanted to make this available to Astra and our shared users.

This will add support for Lessons and Quizzes and if LifterLMS Assignments is enabled it will add the theme fields to assignments as well. I didn't add any dependency checks for assignments as it's all in the filter and the array or data will be ignored if assignments don't exist.

Here's a screencap of what it looks like for a lesson:

course_builder_ _llms-72_dev_ _wordpress

Thanks for the beautiful theme as always and let me know if you have any questions or concerns with this!


Nikschavan:
Hey @thomasplevy,

Thank you very much for this PR and all the other ones that you have sent.
I am waiting on this one to add the similar required changes to Astra Pro's settings and release both of these together.

PR #547, opened 2018-05-01
remove post navigation from LifterLMS Assignments single posts

thomasplevy:
Like #476 this removes post nav from "Assignments" which are available via the LifterLMS Assignments add-on.

This was reported to me as a bug in LifterLMS today and I'd rather have this fixed in Astra core if you're okay with it instead of providing custom code snippets or CSS hacks when this is reported to us as a bug

Thanks as always!


Nikschavan:
Yes, It would be correct o fix this from Astra theme itself. thank you for this @thomasplevy

PR #672, opened 2018-07-20
Checkbox retina logo

Balachandark:
* Added checkbox option for site identity retina logo


inc/extras.php
```diff
@@ -1330,21 +1330,26 @@ function astra_replace_header_attr( $attr, $attachment, $size ) {
 				$attr['class'] = 'astra-logo-svg';
 			}
 
-			$retina_logo = astra_get_option( 'ast-header-retina-logo' );
+			$diff_retina_logo = astra_get_option( 'different-retina-logo' );
 
-			$attr['srcset'] = '';
+			if ( $diff_retina_logo ) {
```
Nikschavan: Perform a strict change here - `'1' === $diff_retina_logo`

inc/theme-update/class-astra-theme-update.php
```diff
@@ -720,6 +724,23 @@ static public function v_1_4_0_beta_5() {
 			update_option( 'astra-settings', $theme_options );
 		}
 
+		/**
+		 * Function to backward compatibility for version less than 1.4.1
```
Nikschavan: Explain in phpdoc what exactly is being migrated and the for a reason.
For eg in this case - 
1. Set the new option `different-retina-logo`  to true for users who are already using a retina logo.

This phpdoc should have all the explanations related to data migrations in this version.

PR #697, opened 2018-07-30
Change default thumbnail size

ipokkel:
Default thumbnail size being pulled for blog archives when set to full results in image size penalties on Google pagespeed for archive pages (below 40%)

For a short term fix the thumbnail size being fetched in "function astra_get_post_thumbnail" was set to large to reach a balance between pagespeed ranking and image quality of post thumbnail being displayed on archive pages.


Nikschavan:
Thank you for this PR @ipokkel

ipokkel:
Thanks for incorporating @Nikschavan

Nikschavan:
I have added this as to our tasks. We will be adding an option to add image size control in the upcoming version of Astra Pro.

PR #705, opened 2018-07-31
Allow new Search Styles from Astra Pro.

Nikschavan:
- Search icon will not turn into a search box when in mobile view.


inc/extras.php
```diff
@@ -277,11 +277,18 @@ function astra_get_dynamic_header_content( $option ) {
...
+		<div class="ast-search-menu-icon slide-search" id="ast-search-form">
+			<div class="ast-search-icon">
+				<a class="slide-search astra-search-icon" href="#">
+					<span class="screen-reader-text"><?php esc_html_e( 'Search', 'astra' ); ?></span>
+				</a>
+			</div>
+			<?php get_search_form(); ?>
+		</div>
+		<?php
+		$search_html = ob_get_clean();
```
Nikschavan: If you use `get_search_form( false )` you should not need object buffering here.

maheshwaghmare: The form that within the `<div class="ast-search-menu-icon ...` so I have kept the form in the buffer.

E.g.
```
</div>
<?php get_search_form(); ?>
</div>
```

PR #737, opened 2018-08-14
Elementor Archive Tempalte styling override by theme's styleing

rushijagani:


inc/compatibility/class-astra-elementor-pro.php
```diff
@@ -262,6 +263,23 @@ public function do_footer() {
 			}
 		}
 
+		/**
+		 * Remove theme post's default classes
```
Nikschavan: ```Remove theme post's default classes when the post is created using Elementor's template builder.```

rushijagani: Function doc is updated

PR #749, opened 2018-08-21
Search style colors

maheshwaghmare:


inc/extras.php
```diff
@@ -347,6 +347,45 @@ function astra_get_search( $option = '' ) {
 	}
 }
 
+/**
+ * Search Form
+ */
+if ( ! function_exists( 'astra_get_custom_html' ) ) :
```
Nikschavan: You need to check for `astra_get_search_form()`

PR #817, opened 2018-09-11
Nav menu css specific

Nikschavan:
Target only `.sub-menu` for hiding the ul container from navigation menu


rushijagani:
When Primary Menu is not selected then default pages set as a menu. So, in this case, the submenu has the `.children` class.

PR #831, opened 2018-09-18
Remove astra wrapper for oembed markup only on admin screen

rushijagani:


inc/class-astra-after-setup-theme.php
```diff
@@ -149,7 +149,9 @@ function setup_theme() {
 				add_editor_style( 'assets/css/' . $dir_name . '/editor-style' . $file_prefix . '.css' );
 			}
 
-			if ( apply_filters( 'astra_fullwidth_oembed', true ) ) {
+			$astra_fullwidth_oembed_on_admin_screen = apply_filters( 'astra_fullwidth_oembed_on_admin_screen', false );
+
+			if ( ( ! is_admin() && apply_filters( 'astra_fullwidth_oembed', true ) ) || ( is_admin() && $astra_fullwidth_oembed_on_admin_screen ) ) {
```
Nikschavan: Instead of using two filters, can we do it by default to admin as well as the frontend in the same filter?

rushijagani: I think we should remove our oembed wrapper by default for admin screen.

PR #877, opened 2018-10-03
replace all insecure links with https

Balachandark:
* Replace all insecure links with https


assets/fonts/astra.svg
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" standalone="no"?>
-<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" >
-<svg xmlns="http://www.w3.org/2000/svg">
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "https://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" >
+<svg xmlns="https://www.w3.org/2000/svg">
 <metadata>Generated by IcoMoon</metadata>
```
Nikschavan: Please verify that fonts work as expected on HTTP and HTTPS sites.
Be sure to clear browser cache when testing.

composer.lock
```diff
@@ -50,12 +50,12 @@
                 {
                     "name": "Franck Nijhof",
```
Nikschavan: This is an autogenerated file and should be left unmodified.

PR #878, opened 2018-10-03
add a new filter to change the post meta separator

Balachandark:
* Added a new filter post_meta_separator to change the post meta separator symbol.


inc/blog/blog-config.php
```diff
@@ -25,6 +25,8 @@ function astra_get_post_meta( $post_meta, $separator = '/' ) {
 		$output_str = '';
 		$loop_count = 1;
 
+		$separator = apply_filters( 'post_meta_separator', $separator );
```
Nikschavan: The filter should have a prefix of `astra_`

PR #880, opened 2018-10-05
Select2 typography enhancement

rushijagani:


inc/customizer/custom-controls/typography/class-astra-control-typography.php
```diff
@@ -190,13 +228,55 @@ protected function render_font( $default ) {
...
+		echo '<label>';
+		$this->render_content_title();
 		echo '</label>';
+		echo '<select ';
+		$this->link();
+		$this->render_connect_attribute();
+		echo ' multiple >';
+		$values = explode( ',', $this->value() );
+		foreach ( $values as $key => $value ) {
+			echo '<option value="' . $value . '" selected="selected" >' . $value . '</option>';
```
Nikschavan: escape `$value` here as attr and html

rushijagani: Added escaping html and attr

PR #908, opened 2018-10-19
Fix: ask for ratings only to the site administrators.

Nikschavan:


Nikschavan:
I believe reviews can be given by the non-administrator users of the site as well. Astra does affect them in the integrations/options provided by the theme and they can share their review as well.

PR #910, opened 2018-10-20
Fix: Gutenberg embeds not responsive

Nikschavan:


inc/class-astra-after-setup-theme.php
```diff
@@ -157,7 +157,6 @@ function setup_theme() {
 			if ( apply_filters( 'astra_fullwidth_oembed', true ) ) {
 				// Filters the oEmbed process to run the responsive_oembed_wrapper() function.
 				add_filter( 'embed_oembed_html', array( $this, 'responsive_oembed_wrapper' ), 10, 3 );
-				add_filter( 'oembed_result', array( $this, 'responsive_oembed_wrapper' ), 10, 3 );
```
Nikschavan: This was added for making the embeds from LearnDash responsive. This should be tested before merginf.

Nikschavan: Tested, works correct!

PR #934, opened 2018-11-29
Add custom menu item button

Balachandark:
* Added button option.


inc/extras.php
```diff
@@ -378,6 +382,40 @@ function astra_get_custom_html( $option_name = '' ) {
...
+		$button_text    = astra_get_option( $button_text );
+		$button_link    = astra_get_option( $button_link );
+		$button_style   = astra_get_option( $button_style );
+		$outside_menu   = astra_get_option( 'header-display-outside-menu' );
+
+		$button_classes = ( 'theme-button' === $button_style ? 'ast-button' : 'ast-custom-button' );
+
+		$outside_menu = apply_filters( 'astra_convert_link_to_button', $outside_menu );
+
+		$custom_html  = '<a class="ast-custom-button-link" href="' . do_shortcode( $button_link ) . '"><button class=' . $button_classes . '>' . do_shortcode( $button_text ) . '</button></a>';
```
Nikschavan: Link, classes, content etc should be escaped appropriately here.

inc/extras.php
```diff
@@ -378,6 +382,40 @@ function astra_get_custom_html( $option_name = '' ) {
...
+		$custom_html .= '<a class="menu-link" href="' . do_shortcode( $button_link ) . '">' . do_shortcode( $button_text ) . '</a>';
```
Nikschavan: This extra button can be avoided in when the button is to be displayed on the desktop as well as mobile layouts. 
You can add this markup conditionally.

inc/extras.php
```diff
@@ -378,6 +382,40 @@ function astra_get_custom_html( $option_name = '' ) {
...
+		$custom_html    = '';
+		$button_classes = '';
+		$button_text    = astra_get_option( $button_text );
+		$button_link    = astra_get_option( $button_link );
+		$button_style   = astra_get_option( $button_style );
+		$outside_menu   = astra_get_option( 'header-display-outside-menu' );
+
+		$button_classes = ( 'theme-button' === $button_style ? 'ast-button' : 'ast-custom-button' );
+
+		$outside_menu = apply_filters( 'astra_convert_link_to_button', $outside_menu );
```
Nikschavan: `$outside_menu` seems unused? Can you check if this is required?

inc/class-astra-dynamic-css.php
```diff
@@ -481,11 +504,101 @@ public static function return_output() {
 				'.ast-header-break-point .main-header-bar .ast-button-wrap .menu-toggle' => array(
 					'border-radius' => ( '' !== $mobile_header_toggle_btn_border_radius ) ? esc_attr( $mobile_header_toggle_btn_border_radius ) . 'px' : '',
 				),
+
```
Nikschavan: Degerate custom CSS for the button only if the button is selected in the custom menu item. This will reduce unused CSS on the page.

inc/extras.php
Balachandark: Links, classes are now properly escaped.

inc/extras.php
Balachandark: Updated the markup.

inc/extras.php
Balachandark: It is now used to add the markup conditionally.

PR #959, opened 2018-12-11
Prevent duplicate lesson navigation templates in LifterLMS 3.25.0 and WP Core 5.0

thomasplevy:
In an effort to improve templating in LifterLMS core LifterLMS templates previously output automatically by action hooks are being transitioned to being output by blocks.

Without this PR, anyone using 3.25.0 of LifterLMS and WP 5.0 (or the Gutenberg plugin and 4.9 or earlier) will see two lesson navigation:

new_lesson_-_llms-72_test

After this update, users will see the template part _only_ if the block has been added to the lesson.

It also ensures that any user who has not yet migrated to 5.0 & 3.25.0 will still have the navigation in the correct area without issue.


thomasplevy:
Thanks @Nikschavan

Nikschavan:
You are welcome @thomasplevy! We greatly appreciate contributions from your team :)

PR #962, opened 2018-12-14
Add widget title filter for WooCommerce and EDD

Balachandark:
* Added filter to change the attributes for EDD and WOO widgets


inc/compatibility/edd/class-astra-edd.php
```diff
@@ -613,25 +613,31 @@ function edd_single_product_class( $classes ) {
...
-				array(
-					'name'          => esc_html__( 'Easy Digital Downloads Sidebar', 'astra' ),
-					'id'            => 'astra-edd-sidebar',
-					'description'   => __( 'This sidebar will be used on Product archive, Cart, Checkout and My Account pages.', 'astra' ),
-					'before_widget' => '<div id="%1$s" class="widget %2$s">',
-					'after_widget'  => '</div>',
-					'before_title'  => '<h2 class="widget-title">',
-					'after_title'   => '</h2>',
+				apply_filters(
+					'astra_edd_downloads_widgets_init',
```
Nikschavan: @Balachandark We should keep the filter name similar to the id of the sidebar. For eg, this should be `astra_edd_sidebar_init`

Balachandark: Filter name changed.

PR #977, opened 2019-01-04
add LifterLMS lesson page builder support

thomasplevy:
After adding support for the block editor in the LifterLMS core we realized that in doing so we've caused new compatibility issues with every page builder in existence as well as the Classic Editor plugin. We've been fixing this by adding checks in the LifterLMS core to determine how the post should be handled but since this all relies on a filter Astra no longer "works" properly on lessons due to the movement of this action in question.

By applying the filters we use in the core it will ensure that whatever we do will also work from within Astra.

I feel a bit silly for submitting another PR about (essentially) the same issue I submitted about a few weeks ago but I can't find another way around it.

Any questions let me know and thanks as always!


Nikschavan:
Thank you for this change @thomasplevy 👍

thomasplevy:
thanks as always folks!

PR #1000, opened 2019-01-23
added read time option for single blog

UttamSharma007:
Added Read title option in the array for a single blog.


Nikschavan:
@UttamSharma007 This option should go in the Astra Pro.

PR #1004, opened 2019-01-25
[Fix] -> Check if default wordpress menu is enabled do not add class …

premanshup:
…ast-mega-menu-enabled. Since this class was overriding the css of default menu arrows.


Nikschavan:
Instead of changing the filter to be conditional, We should add the classes to this filter from the Astra Pro conditionally.

PR #1051, opened 2019-03-04
[Fix] -> When title and meta option is off from customizer for single…

premanshup:
… blog, the margin was appearing which was creating spacing. Added condition to replace the class.


inc/extras.php
```diff
@@ -1296,13 +1296,16 @@ function astra_get_footer_widget( $sidebar_id ) {
...
-
-		if ( empty( $title_markup ) && empty( $thumb_markup ) && ( is_page() || empty( $post_meta_markup ) ) ) {
+		$post_id                    = astra_get_post_id();
+		$classes                    = array();
+		$title_markup               = astra_the_title( '', '', $post_id, false );
+		$thumb_markup               = astra_get_post_thumbnail( '', '', false );
+		$post_meta_markup           = astra_single_get_post_meta( '', '', false );
+		$blog_single_post_structure = astra_get_option( 'blog-single-post-structure' );
+
+		if ( ! $blog_single_post_structure || ( 'single-image' === astra_get_prop( $blog_single_post_structure, 0 ) && empty( $thumb_markup ) ) ) {
```
premanshup: 1. In the earlier condition, the Global customizer option to deactivate the Title meta was not checked due to which the classes ast-no-title & ast-no-thumbnail were getting added to the HTML.

Now, checked it using $blog_single_post_structure variable.

2. OR condition for single-image in the `if`:
- When Featured Image global condition was enabled and no image was added from Edit Post in featured image for a post, the class ast-no-thumbnail was getting added.

Checked the condition, If global Featured Image is enabled, then check if the featured image is set or not, if not then do not add class ast-no-thumbnail due to which margin-bottom gets added.

PR #1066, opened 2019-03-07
AMP Compatibility: Fix usage of amp-state component

westonruter:
I noticed quite a few AMP validation errors when testing the theme, and they turned out to just be due to `<amp-state>` being incorrectly placed in the `head` when it has to be in the `body`.

Also, these attributes on `<amp-state>` are added automatically by the `amp-state` component and they should be omitted:

* `class="i-amphtml-element i-amphtml-layout-container"`
* `hidden=""` 
* `aria-hidden="true"`


inc/compatibility/class-astra-amp.php
```diff
@@ -1013,7 +1013,7 @@ public function nav_menu_wrapper( $attr ) {
 		 * @return String HTML MArkup for the menu including the AML State.
 		 */
 		public function toggle_button_markup( $item_output, $item ) {
-			$item_output .= '<amp-state id="neveMenuItemExpanded' . esc_attr( $item->ID ) . '" class="i-amphtml-element i-amphtml-layout-container" hidden="" aria-hidden="true"><script type="application/json">false</script></amp-state>';
+			$item_output .= '<amp-state id="neveMenuItemExpanded' . esc_attr( $item->ID ) . '"><script type="application/json">false</script></amp-state>';
```
westonruter: Is this the `astra` theme or the `neve` theme? 😉

Nikschavan:
Thank you very much for this @westonruter, it is highly appreciated!

PR #1136, opened 2019-04-05
Improvement: Add font-display property for the Astra Fonts

Nikschavan:
https://css-tricks.com/really-dislike-fout-font-display-optional-might-jam/


gravnetic:
When I test my site with Astra enabled all font variants (woff2, off, ttf, eot, svg) are downloaded. 
 This is not a desirable solution.  I am optimizing my site and would prefer to have the highest performing compatible font load.

https://www.filamentgroup.com/lab/font-loading.html
https://www.freecodecamp.org/news/web-fonts-in-2018-f191a48367e8/

Please advise, and if support declaration should be added to the @font-face code.  Also, I would like to see a defer or async option added to all Astra assets like has been done with fonts.

PR #1142, opened 2019-04-12
[Fix] -> Removed extra CSS selector from menu li

premanshup:


Nikschavan:
@premanshup Can you verify once if this is to be merged as this is an old PR?

premanshup:
@Nikschavan Sir / @patilvikasj I have cross-checked this and found that we can close this PR. If we remove the selectors, the css applies to above header & below header menus as well which conflicts with the megamenus in them. For now, we can close this PR.

PR #1163, opened 2019-05-04
Allows custom Wordpress apps to set the title.

kharysharpe:
Added 'astra_the_title' filter to allow overriding the title.

Whenever you are embedding a custom app (think Laravel like app) within Wordpress which manages its own route/url and page rendering, this allows those non Wordpress rendered pages to set the 'page' title.

add_filter('astra_the_title', function($title)
{
    return 'My App';
});


Nikschavan:
Thank you for the PR @kharysharpe 👍

kharysharpe:
Hey @Nikschavan,

I was hoping this would have landed in 1.8.3, seems like it missed being merged.

Can this be merged now in preparation for the next release?

-- Khary

PR #1202, opened 2019-06-05
version 1.8.5 Update

patilvikasj:


changelog.txt
```diff
@@ -1,3 +1,11 @@
+v1.8.5
```
premanshup: Need to add  '-'  at the beginning like other changelogs.

patilvikasj: > Need to add '-' at the beginning like other changelogs.

Done.

PR #1234, opened 2019-06-24
Version 2.0.0

Nikschavan:
Version 2.0.0-rc.1

- Improvement: Improved logic that is used to save the values of options inside the new popups, Earlier all the values of options inside the popup were combined and stored in JSON objects now all the values inside the popup are stored separately similar to Astra Version 1.X.X. This removes the requirement of the Database Upgrade routine when updating from Version 1.X to 2.X
- Improvement: Changed the images inside customizer to new lightweight SVGs.
- Improvement: Updated minor UI tweaks for all the customizer controls to unify the design across all the options.


inc/theme-update/class-astra-theme-update.php
```diff
@@ -195,6 +224,10 @@ static public function init() {
 				self::v_1_6_1();
 			}
 
+			if ( version_compare( $saved_version, '2.0.0', '<' ) ) {
```
Nikschavan: @Balachandark  The version number here should be 2.0.0-beta.1

PR #1240, opened 2019-06-25
refactor: batch processing

Balachandark:
* Added background updater class for processing all the future DB migrations functions in batches.


inc/theme-update/class-astra-theme-update.php
```diff
@@ -1120,6 +1038,25 @@ public static function individual_queued_item_operations( $group ) {
 
 			return $new_options;
 		}
+
+		/**
+		 * Customizer Version 2 categorization database opertions using batch processing
+		 *
+		 * @return void
+		 */
+		public static function astra_theme_update_v2_0_0_customizer_optimization() {
```
patilvikasj: This function can go in a separate class for this specific update migration task. This will keep every update migration process separated by individual class.

inc/theme-update/class-astra-theme-batch-update.php
```diff
@@ -0,0 +1,200 @@
...
+		 */
+		private static $background_updater;
+
+		/**
+		 * DB updates and callbacks that need to be run per version.
+		 *
+		 * @var array
+		 */
+		private static $db_updates = array(
+			'2.0.0-beta.1' => array(
```
patilvikasj: This array will contain class-specific methods instead of the static method from the same class.

inc/theme-update/astra-update-functions.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+}
+
+/**
+ * Added a JSON object for changes in customizer version 2 database migration.
+ *
+ * @since x.x.x
+ * @return array
+ */
+function astra_new_controls() {
```
Nikschavan: keep a prefix of version number for this function as well. `astra_theme_update_v2_0_0_new_controls`

inc/theme-update/astra-update-functions.php
```diff
@@ -0,0 +1,111 @@
...
+ * @param array $new_options New options data.
+ * @return void
+ */
+function control_data_update( $new_options ) {
+
+	$theme_options = get_option( 'astra-settings', array() );
+
+	$theme_options = array_merge( $theme_options, $new_options );
+
+	update_option( 'astra-settings', $theme_options );
```
Nikschavan: `control_data_update` is being run in `foreach` so basically `update_option` and `get_option` are being called in the loop, We should about this.

You can simply avoid this function and add this code inline and run `get_option` and `update_option` out of the loop.

inc/theme-update/astra-update-functions.php
```diff
@@ -0,0 +1,111 @@
...
+	$new_options = array();
+
+	$json = astra_new_controls();
+
+	$json_array = json_decode( $json, true );
+
+	$sub_control = $json_array[ $group ];
+
+	// Check if group key exists in the theme options.
+	if ( array_key_exists( $group, $theme_options ) ) {
```
Nikschavan: You can do this check in `astra_theme_update_v2_0_0_customizer_optimization()` and avoid this function call alltogether.

inc/theme-update/astra-update-functions.php
```diff
@@ -0,0 +1,111 @@
...
+ * Perform database operations on indivitual group controls
+ *
+ * @since x.x.x
+ *
+ * @param string $group Queue item.
+ * @return array
+ */
+function group_item_operations( $group ) {
+
+	$theme_options = get_option( 'astra-settings', array() );
```
Nikschavan: Can these options be passed as input argument from `astra_theme_update_v2_0_0_customizer_optimization()`

inc/theme-update/astra-update-functions.php
```diff
@@ -0,0 +1,111 @@
...
+
+	$new_options = array();
+
+	$json = astra_new_controls();
```
Nikschavan: Can these options be passed as input argument from `astra_theme_update_v2_0_0_customizer_optimization()`

Nikschavan: instead of sending complete `astra_new_controls()` to this function you can send relavant `$group` for processing here.

inc/theme-update/astra-update-functions.php
Balachandark: function name changed to astra_theme_update_v2_0_0_new_controls

inc/theme-update/astra-update-functions.php
Balachandark: The $group and $sub_control are now passed from `astra_theme_update_v2_0_0_customizer_optimization` function so the JSON won't be required in this function.

inc/theme-update/astra-update-functions.php
Balachandark: astra-settings are now passed from the `astra_theme_update_v2_0_0_customizer_optimization` function

inc/theme-update/astra-update-functions.php
Balachandark: `array_key_exists` has now been checked in `astra_theme_update_v2_0_0_customizer_optimization` function. so the `group_item_operations()` won't run untill the key exists.

inc/theme-update/astra-update-functions.php
Balachandark: I have changed the logic. Now update_option will run just once after the loop is executed.

inc/theme-update/batch-processing/class-wp-background-process-astra-theme.php
```diff
@@ -0,0 +1,81 @@
...
+		 * @var string
+		 */
+		protected $action = 'database_migration';
+
+		/**
+		 * Really long running process
+		 *
+		 * @return int
+		 */
+		public function really_long_running_task() {
```
Nikschavan: We can get rid of this method.

Balachandark: Yes, I have now removed this function

inc/theme-update/class-astra-theme-background-updater.php
```diff
@@ -0,0 +1,191 @@
...
+				if ( version_compare( $current_db_version, $version, '<' ) ) {
+					foreach ( $update_callbacks as $update_callback ) {
+						error_log( sprintf( 'Queuing %s - %s', $version, $update_callback ) );
+
+						self::$background_updater->push_to_queue( $update_callback );
+					}
+				}
+			}
+
+			self::$background_updater->push_to_queue( $this->update_db_version() );
```
patilvikasj: Try to pass task parameter here with the function name. like `push_to_queue( array( 'task' => 'trash_pending_orders' ) );`

inc/theme-update/class-astra-theme-background-updater.php
```diff
@@ -0,0 +1,196 @@
...
+			}
+		}
+
+		/**
+		 * Is this a brand new theme install?
+		 *
+		 * @since x.x.x
+		 * @return boolean
+		 */
+		private function is_new_install() {
```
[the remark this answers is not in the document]
Balachandark: I have moved the `is_new_install` function to the common functions file.
Also, as discussed as the data is added to the theme mod so there won't be additional database calls so we will keep the astra_get_options as is.

inc/theme-update/batch-processing/class-wp-async-request.php
```diff
@@ -0,0 +1,163 @@
+<?php
```
Nikschavan: Should the batch-processing library be moved to the [`/lib`](https://github.com/brainstormforce/astra/tree/master/inc/lib) directory?
That way entire `/lib` can be excluded from `phpcs`

inc/theme-update/batch-processing/class-wp-background-process-astra-theme.php
```diff
@@ -0,0 +1,69 @@
...
+			return false;
+		}
+
+		/**
+		 * Complete
+		 *
+		 * Override if applicable, but ensure that the below actions are
+		 * performed, or, call parent::complete().
+		 *
+		 * @since 2.0.0
```
Nikschavan: This since tag will need to be updated.

inc/theme-update/class-astra-theme-background-updater.php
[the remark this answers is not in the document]
premanshup: As per the discussion, We won't be needing this function globally so we can keep it here only.

inc/theme-update/batch-processing/class-wp-background-process-astra-theme.php
Balachandark: Changed the tag to x.x.x

cghooks.lock
```diff
@@ -0,0 +1 @@
+["pre-commit"]
```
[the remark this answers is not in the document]
Balachandark: Yes, I have updated the branch and this file has been ignored from `.gitignore`.

inc/theme-update/batch-processing/class-wp-async-request.php
Balachandark: I have moved the code to the lib folder

PR #1324, opened 2019-07-18
Sub control save logic

patilvikasj:


inc/customizer/class-astra-customizer.php
```diff
@@ -274,6 +283,51 @@ private function register_section( $config, $wp_customize ) {
...
+		 * @param Array                $control_config Panel Configuration settings.
+		 * @param WP_Customize_Manager $wp_customize instance of WP_Customize_Manager.
+		 * @since 1.4.3
+		 * @return void
+		 */
+		private function register_sub_control_setting( $control_config, $wp_customize ) {
+
+			$sub_control_name = ASTRA_THEME_SETTINGS . '[' . astra_get_prop( $control_config, 'name' ) . ']';
+
+			$section = ( astra_get_prop( $control_config, 'section' ) ) ? astra_get_prop( $control_config, 'section' ) : 'title_tagline';
```
Nikschavan: `astra_get_prop` already supports third parameter for default value, so you can use `astra_get_prop( $control_config, 'section', 'title_tagline' ) ` instead of using the terniary condition here.

inc/customizer/custom-controls/hidden/class-astra-control-hidden.php
```diff
@@ -0,0 +1,72 @@
+<?php
+/**
+ * Customizer Control: divider
+ *
+ * @package     Astra
+ * @author      Astra
+ * @copyright   Copyright (c) 2019, Astra
+ * @link        https://wpastra.com/
+ * @since       1.0.0
```
Nikschavan: Can you change @sine docBlock to `x.x.x`

PR #1358, opened 2019-07-29
Css file generation

Balachandark:


inc/addons/breadcrumbs/dynamic-css/dynamic.css.php
```diff
@@ -59,8 +63,7 @@ function astra_breadcrumb_section_dynamic_css( $dynamic_css, $dynamic_css_filter
 	$breadcrumb_position = astra_get_option( 'breadcrumb-position', 'none' );
```
patilvikasj: You can remove this call and check below that.

inc/assets/js/astra-admin-menu-settings.js
```diff
@@ -245,6 +245,35 @@
...
 
 		},
+
+		_astra_refresh_assets: function(e)
+		{
+			var _this = $(this)
+
+			_this.addClass('refreshing-assets');
+			_this.find( '.ast-loader' ).addClass( 'spinner is-active' );
+			_this.find( '.ast-refresh-btn-text' ).text('Refreshing...');
```
patilvikasj: Can we have this text translatable?

inc/class-astra-helper.php
```diff
@@ -0,0 +1,361 @@
...
+		 * @return void
+		 */
+		public function astra_refresh_assets() {
+
+			astra_delete_option( 'file-write-access' );
+
+			$uploads_dir      = $this->astra_get_upload_dir();
+			$uploads_dir_path = $uploads_dir['path'];
+
+			array_map( 'unlink', glob( $uploads_dir_path . '/astra-*.*' ) );
```
patilvikasj: Can we delete only specific assets files? maybe we can have a folder name accordingly. astra-assets or something.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,361 @@
...
+			if ( ! empty( $style_data ) ) {
+				$this->file_write( $style_data, $slug, $timestamp, $type, $assets_info );
+			}
+
+			$uploads_dir     = $this->astra_get_upload_dir();
+			$uploads_dir_url = $uploads_dir['url'];
+
+			$write_access = astra_get_option( 'file-write-access', true );
+
+			if ( ! $write_access ) {
```
patilvikasj: Can we have a filter that will disable the loading of separate CSS?

inc/class-astra-helper.php
```diff
@@ -0,0 +1,361 @@
...
+					return;
+				}
+			}
+
+			// Create a new file.
+			$put_contents = $wp_filesystem->put_contents( $assets_info['path'], $style_data );
+
+			astra_update_option( 'file-write-access', $put_contents );
+
+			if ( false === $this->astra_get_archive_title() ) {
```
patilvikasj: Check if we can take value astra_get_archive_title function returns in static variable? That will reduce the number of calls and functions astra_get_archive_title function internally makes?

inc/addons/breadcrumbs/dynamic-css/dynamic.css.php
Balachandark: I have already checked the condition at the start of the function so now I have removed the duplicate condition check.

inc/assets/js/astra-admin-menu-settings.js
Balachandark: I have now made the text translatable.

inc/class-astra-helper.php
Balachandark: I have now changed the code as follows so as to delete only the theme and addon css files
```
array_map( 'unlink', glob( $uploads_dir_path . '/astra-theme-dynamic-css*.*' ) );
array_map( 'unlink', glob( $uploads_dir_path . '/astra-addon-dynamic-css*.*' ) );
```

inc/class-astra-helper.php
Balachandark: added a new filter for it `astra_load_dynamic_css_inline`

inc/class-astra-helper.php
Balachandark: I have not used a static variable but have passed the values as parameters to other functions so as to reduce the `astra_get_archive_title ` function calls.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+			return $post_id;
+		}
+
+		/**
+		 * Gets the current timestamp.
+		 *
+		 * @since x.x.x
+		 * @return string $timestamp Timestamp.
+		 */
+		public function astra_get_file_timestamp() {
```
Nikschavan: You can avoid prefix `astra_` from method names.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+			}
+		}
+
+		/**
+		 * Gets the current post id.
+		 *
+		 * @since x.x.x
+		 * @return string $post_id Post ID.
+		 */
+		public function astra_get_post_id() {
```
Nikschavan: You can avoid prefix astra_ from method names.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+
+			global $wp_filesystem;
+
+			$wp_info  = wp_upload_dir( null, false );
+			$dir_name = basename( ASTRA_THEME_DIR );
+			if ( 'astra' == $dir_name ) {
+				$dir_name = 'astra';
+			}
+			// SSL workaround.
+			if ( $this->is_ssl() ) {
```
Nikschavan: Any reason we cannot _just_ use [`is_ssl`](https://developer.wordpress.org/reference/functions/is_ssl/) here and have to check other headers as well?

inc/class-astra-helper.php
```diff
@@ -0,0 +1,335 @@
...
+			} else {
+				$post_timestamp = get_option( 'astra_' . $type . '_get_dynamic_css' );
+			}
+
+			if ( '' == $post_timestamp || ! file_exists( $assets_info['path'] ) ) {
+				$timestamp = $this->astra_get_file_timestamp();
+			} else {
+				$timestamp = $post_timestamp;
+			}
+
```
Nikschavan: All the above conditions to `get_the_timestamp` can be divided into a separate method.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,335 @@
...
+		 *
+		 * @param  string $style_data   Gets the CSS data.
+		 * @param  string $type         Gets the type theme/addon.
+		 * @since  x.x.x
+		 * @return void
+		 */
+		public function enqueue_styles( $style_data, $type ) {
+
+			$slug          = $this->astra_get_archive_title();
+			$archive_title = $this->astra_get_archive_title();
```
Nikschavan: If both the variables hold the same value, can these be combined into a single variable?

inc/class-astra-helper.php
```diff
@@ -0,0 +1,335 @@
...
+			// Adds an option as if the uploads folder has file rights access.
+			astra_update_option( 'file-write-access', $put_contents );
+
+			if ( false === $archive_title ) {
+				// Update the post meta.
+				update_post_meta( get_the_ID(), 'astra_' . $type . '_style_timestamp_css', $timestamp );
+			} else {
+				// Update the option.
+				update_option( 'astra_' . $type . '_get_dynamic_css', $timestamp );
+			}
```
Nikschavan: These `if/else` blocks can be moved to a separate `update_timestamp` method.

inc/core/class-astra-admin-settings.php
```diff
@@ -739,6 +742,62 @@ public static function astra_welcome_page_five_star_scetion() {
...
+			<div class="postbox">
+				<h2 class="hndle ast-normal-cusror">
+					<span class="dashicons dashicons-update"></span>
+					<span>
+						<?php printf( esc_html( 'Refresh Assets file', 'astra' ) ); ?>
+					</span>
+				</h2>
+				<div class="inside">
+					<p class="warning">
+						<?php echo $message; ?>
```
Nikschavan: `$message` should be escaped.

inc/core/class-astra-admin-settings.php
```diff
@@ -739,6 +742,62 @@ public static function astra_welcome_page_five_star_scetion() {
...
+					</p>
+					<p>
+					<?php
+						$a_tag_open  = '<a target="_blank" rel="noopener" href="' . $doc_link . '">';
```
Nikschavan: `$doc_link` should be escaped.

inc/core/class-astra-admin-settings.php
```diff
@@ -739,6 +742,62 @@ public static function astra_welcome_page_five_star_scetion() {
...
+							$a_tag_open,
+							$a_tag_close
+						);
+					?>
+					</p>
+
+					<label for="astra_refresh_assets">
+						<button class="button astra-refresh-assets" id="astra_refresh_assets">
+							<span class="ast-loader"></span>
+							<span class="ast-refresh-btn-text"><?php echo $button_text; ?></span>
```
Nikschavan: `$button_text` should be escaped.

inc/core/class-astra-admin-settings.php
Balachandark: already did use esc_url while defining the variable.

inc/core/class-astra-admin-settings.php
Balachandark: already did use esc_html__ while defining the variable.

inc/core/class-astra-admin-settings.php
Balachandark: $button_text is now escaped while defining the varaible

inc/class-astra-helper.php
Balachandark: Added a new function `update_timestamp` to carry out the update post meta / option method.

inc/class-astra-helper.php
Balachandark: We modify the $slug variable in the function so we need both the values i.e the modified $slug and the original value that would be available in $archive_title

Balachandark: Yes, I have now combined this into a single variable

inc/class-astra-helper.php
Balachandark: I have added two new functions for it
`astra_get_post_timestamp` -> Returns the current Post Meta/ Option Timestamp.
`astra_maybe_get_new_timestamp` -> Returns the current Post Meta/ Option Timestamp or creates a new timestamp.

inc/class-astra-helper.php
Balachandark: There are two more conditions that we need to check that the default `is_ssl()` does not check.
i.e 
1) get_option( 'siteurl' ), 'https://' )
2) $_SERVER['HTTP_X_FORWARDED_PROTO']

So I have added a new function for it.

inc/class-astra-helper.php
Balachandark: I have now removed the `astra_` prefix from most of the functions.

inc/class-astra-helper.php
Balachandark: I have now removed the astra_ prefix from most of the functions.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+
+		/**
+		 * Enqueue theme CSS files.
+		 *
+		 * @since x.x.x
+		 * @return void
+		 */
+		public function theme_enqueue_styles() {
+
+			$slug = $archive_title = $this->astra_get_archive_title();
```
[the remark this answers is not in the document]
Balachandark: I have now changed this to use only one variable.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+		}
+
+		/**
+		 * Remove post meta that check if css file need to be regenerated.
+		 *
+		 * @param int $post_id Gets the post id.
+		 * @since x.x.x
+		 * @return void
+		 */
+		public function check_values( $post_id ) {
```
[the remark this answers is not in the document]
Balachandark: maybe we can use `refresh_post_meta_data()` rather than `referesh_post_assets()`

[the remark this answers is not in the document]
Balachandark: I have now added `refresh_post_meta_data`. Please let me know if I need to change it

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+		 */
+		public function astra_refresh_assets() {
+
+			astra_delete_option( 'file-write-access' );
+
+			$uploads_dir      = $this->astra_get_upload_dir();
+			$uploads_dir_path = $uploads_dir['path'];
+
+			array_map( 'unlink', glob( $uploads_dir_path . '/astra-theme-dynamic-css*.*' ) );
+			array_map( 'unlink', glob( $uploads_dir_path . '/astra-addon-dynamic-css*.*' ) );
```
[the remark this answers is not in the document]
Balachandark: We cannot use `wp_delete_file()` as we want to delete all the files from the folder that match a particular pattern. Passing a pattern is not possible. Even if we do so using functions such as glob then it returns an array and the condition will fail.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+
+			if ( defined( 'ASTRA_EXT_FILE' ) ) {
+				add_action( 'wp_enqueue_scripts', array( $this, 'addon_enqueue_styles' ), 999 );
+			}
+
+			add_action( 'astra_post_meta_updated', array( $this, 'check_values' ), 10, 1 );
+
+			// Refresh assets.
+			add_action( 'customize_save_after', array( $this, 'astra_refresh_assets' ) );
+			add_action( 'astra_addon_activate', array( $this, 'astra_refresh_assets' ) );
```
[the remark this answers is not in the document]
Balachandark: I have moved this code to astra addon

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+				require_once ABSPATH . 'wp-admin/includes/file.php';
+			}
+
+			WP_Filesystem();
+
+			add_action( 'wp_enqueue_scripts', array( $this, 'theme_enqueue_styles' ), 1 );
+
+			if ( defined( 'ASTRA_EXT_FILE' ) ) {
+				add_action( 'wp_enqueue_scripts', array( $this, 'addon_enqueue_styles' ), 999 );
+			}
```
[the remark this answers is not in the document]
Balachandark: I have moved this code to astra addon

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+ * Astra Helper.
+ *
+ * @package Astra
+ */
+
+if ( ! class_exists( 'Astra_Helper' ) ) {
+	/**
+	 * Class Astra_Helper.
+	 */
+	class Astra_Helper {
```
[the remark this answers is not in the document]
Balachandark: `Astra_Cache` sounds good to me. I have renamed the class name as well as the file name.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+
+		/**
+		 * Get fonts to generate.
+		 *
+		 * @since x.x.x
+		 * @var array $fonts
+		 */
+		static private $css = array();
```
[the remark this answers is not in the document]
Balachandark: This static variable was not used elsewhere in the code so I have now removed it.

inc/class-astra-helper.php
```diff
@@ -0,0 +1,366 @@
...
+		/**
+		 *  Constructor
+		 */
+		public function __construct() {
+
+			if ( ! function_exists( 'WP_Filesystem' ) ) {
+				require_once ABSPATH . 'wp-admin/includes/file.php';
+			}
+
+			WP_Filesystem();
```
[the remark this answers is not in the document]
Balachandark: Create a new class for it and extended the same in Astra_Cache class

PR #1382, opened 2019-08-09
fix: Introducing getter for Theme Options Page Slug.

premanshup:


inc/core/class-astra-admin-settings.php
```diff
@@ -126,6 +126,26 @@ public static function init_admin_settings() {
...
 
+		/**
+		 * Plugin Slug setter including White Label.
+		 *
+		 * @since x.x.x
+		 * @return void
+		 */
+		public static function astra_set_plugin_slug() {
+			self::$plugin_slug = apply_filters( 'astra_theme_page_slug', self::$plugin_slug );
+		}
```
Nikschavan: A "Setter" should not be required as we are simply applying the filter, you can just keep the `get_theme_page_slug`  method which will `apply_filter` and return the values.

inc/core/class-astra-admin-settings.php
```diff
@@ -126,6 +126,26 @@ public static function init_admin_settings() {
...
+
+		/**
+		 * Plugin Slug getter including White Label.
+		 *
+		 * @since x.x.x
+		 * @return string Plugin Slug.
+		 */
+		public static function astra_get_plugin_slug() {
```
Nikschavan: Should the name of the function be `get_theme_page_slug` as it is mainly for the theme page? or it is also used by other plugins as well?

premanshup: It is used by Astra Pro plugin but ultimately it returns the Astra Options page slug, so yes we can rename it to `astra_get_theme_page_slug()`

inc/core/class-astra-admin-settings.php
premanshup: It can be done but for better practice, I was adding setter and getter functions for the Theme page slug. Please let me know your thoughts.

inc/core/class-astra-admin-settings.php
Nikschavan: Ok, let's change the name then. Also, drop the `astra_` from the name as it already present in the class prefix.

inc/core/class-astra-admin-settings.php
Nikschavan: My opinion is to drop the setter in this case.

premanshup: Okay! I will make changes right away.

premanshup: Changes done!

inc/core/class-astra-admin-settings.php
premanshup: Changes done!

inc/core/class-astra-admin-settings.php
premanshup: Moved `apply_filters` inside get_theme_page_slug function.

PR #1389, opened 2019-08-13
Displaying notice for minimum version for Astra Pro addon

patilvikasj:


inc/core/class-astra-admin-settings.php
```diff
@@ -248,6 +251,42 @@ public static function register_notices() {
...
 
+		/**
+		 * Display notice for minimun version for Astra addon.
+		 *
+		 * @since 2.0.0
+		 */
+		public static function minimum_addon_version_notice() {
+
+			$astra_theme_name = 'Astra';
+			if ( function_exists( 'astra_get_theme_name' ) ) {
```
Nikschavan: Is this `function_exists ` check required as this function is also defined in the theme as well?

inc/core/class-astra-admin-settings.php
```diff
@@ -248,6 +251,42 @@ public static function register_notices() {
...
+				'type'           => '',
+				'message'        => sprintf(
+					/* translators: %1$1s: Theme Name, %2$2s: Minimum Required version of the addon */
+					__( 'Glad to see you have updated the %1$1s theme! Please update the %1$1s Pro addon to version %2$2s or higher.', 'astra' ),
+					$astra_theme_name,
+					ASTRA_EXT_MIN_VER
+				),
+				'priority'       => 1,
+				'type'           => 'warning',
+				'show_if'        => $show_if,
```
Nikschavan: You can also add the condition here instead of defining it outside in the variable `$show_if`

inc/lib/notices/class-astra-notices.php
```diff
@@ -250,7 +250,12 @@ public static function markup( $notice = array() ) {
 		 * @return array       Notice wrapper classes.
 		 */
 		private static function get_wrap_classes( $notice ) {
-			$classes   = array( 'astra-notice', 'notice', 'is-dismissible' );
+			$classes = array( 'astra-notice', 'notice' );
+
+			if ( ! isset( $notice['is_dismissible'] ) || ( isset( $notice['is_dismissible'] ) && $notice['is_dismissible'] ) ) {
```
Nikschavan: Instead of complicating the `is_dismissible` condition here for maintaining the bc, you can define this attribute as default true [in the default's array](https://github.com/brainstormforce/astra/blob/76a3f58fff36a1079ea42488b8d0f57763a538e4/inc/lib/notices/class-astra-notices.php#L180).

That way you will just have to check `true/false` for this attribute and add the class is set to true.

PR #1402, opened 2019-08-19
fix: Omit printing skip-focus-fix script on AMP content

amedina:
Updated to the latest version of the Astra theme and encountered an AMP validation error (see attachment) 
Screen Shot 2019-08-19 at 2 37 58 PM

This validation error can be fixed by preventing the printing of the `skip-focus-fix` script, since it is covered by the AMP framework.

See: https://github.com/ampproject/amphtml/issues/18671


Nikschavan:
Thank you very much for the fix @amedina!

PR #1406, opened 2019-08-20
refactor: use astra_attr to build markup where schema tags are added.

Nikschavan:
`astra_attr()` allows to filter the markup and remove the attributes.

### DIFF
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -589,23 +589,26 @@ if ( ! function_exists( 'astra_header_markup' ) ) {
 	function astra_header_markup() {
 
 		do_action( 'astra_header_markup_before' );
-
-		echo '<header ';
-
-			echo astra_attr(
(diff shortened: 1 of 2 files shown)


inc/extras.php
```diff
@@ -589,9 +589,23 @@ function astra_get_small_footer_menu() {
...
+					'itemscope' => 'itemscope',
+					'id'        => 'masthead',
+					'class'     => join( ' ', astra_get_header_classes() ),
+					'role'      => 'banner',
+				)
+			);
+
+		echo '>';
+
+		?>
```
premanshup: There is no need for closing and opening the PHP tag since most of the below lines are PHP code.

Nikschavan: @premanshup I modified this slightly, I am keeping the tag `<header>` in HTML

PR #1428, opened 2019-08-30
fix: Load CSS JS on the basis of filter to disable enqueueing of assets.

premanshup:


inc/core/class-astra-enqueue-scripts.php
```diff
@@ -287,6 +285,16 @@ public function gutenberg_assets() {
 			wp_add_inline_style( 'astra-block-editor-styles', apply_filters( 'astra_block_editor_dynamic_css', Gutenberg_Editor_CSS::get_css() ) );
 		}
 
+		/**
+		 * Function to check if enqueuing of Astra assets are disabled.
+		 *
+		 * @since x.x.x
+		 * @return boolean
+		 */
+		public static function astra_enqueue_theme_assets() {
```
Nikschavan:
```suggestion
		public static function enqueue_theme_assets() {
```

inc/core/class-astra-enqueue-scripts.php
```diff
@@ -149,9 +149,7 @@ public function add_fonts() {
 		 */
 		public function enqueue_scripts() {
 
-			$astra_enqueue = apply_filters( 'astra_enqueue_theme_assets', true );
-
-			if ( ! $astra_enqueue ) {
+			if ( ! Astra_Enqueue_Scripts::astra_enqueue_theme_assets() ) {
```
Nikschavan:
```suggestion
		public static function enqueue_theme_assets() {
```

inc/cache/class-astra-cache.php
```diff
@@ -97,8 +97,11 @@ public function setup_cache() {
 			$this->write_assets( $theme_css_data, 'theme' );
 		}
 
-		// Call enqueue styles function.
-		$this->enqueue_styles( 'theme' );
+		if ( Astra_Enqueue_Scripts::astra_enqueue_theme_assets() ) {
```
Nikschavan:
```suggestion
		public static function enqueue_theme_assets() {
```

Nikschavan: We can drop `astra` from the method name as the slug is already present in the class name.

PR #1441, opened 2019-09-04
Refactor: Added edit icons in customizer

Balachandark:


inc/customizer/class-astra-customizer.php
```diff
@@ -100,6 +100,7 @@ public function __construct() {
 			add_action( 'customize_register', array( $this, 'customize_register_panel' ), 2 );
 			add_action( 'customize_register', array( $this, 'customize_register' ) );
 			add_action( 'customize_save_after', array( $this, 'customize_save' ) );
+			add_action( 'wp_head', array( $this, 'preview_styles' ) );
```
Balachandark: You have added the internal CSS in this file as well as in here - https://github.com/brainstormforce/astra/pull/1441/files#diff-247790e22304b10f9fc013afc11cd66cR47

We should add both the CSS in the same function.

PR #1445, opened 2019-09-07
fix: Header merge case with last menu item

UttamSharma007:
Merge menu case not working with hiding last menu item


inc/extras.php
```diff
@@ -666,11 +666,19 @@ function astra_toggle_buttons_markup() {
 		$disable_primary_navigation = astra_get_option( 'disable-primary-nav' );
 		$custom_header_section      = astra_get_option( 'header-main-rt-section' );
 		$hide_custom_menu_mobile    = astra_get_option( 'hide-custom-menu-mobile', false );
+		$above_header_merge         = astra_get_option( 'above-header-merge-menu' );
+		$above_header_on_mobile     = astra_get_option( 'above-header-on-mobile' );
+		$below_header_merge         = astra_get_option( 'below-header-merge-menu' );
```
premanshup: @UttamSharma007 We are not using this variable anywhere, can we please remove this?

inc/extras.php
```diff
@@ -666,11 +666,19 @@ function astra_toggle_buttons_markup() {
...
+		$below_header_on_mobile     = astra_get_option( 'below-header-on-mobile' );
 		$menu_bottons               = true;
 
 		if ( ( $disable_primary_navigation && 'none' == $custom_header_section ) || ( $disable_primary_navigation && true == $hide_custom_menu_mobile ) ) {
 			$menu_bottons = false;
+			if ( ( true == $above_header_on_mobile && true == $above_header_merge ) || ( true == $above_header_on_mobile && true == $below_header_on_mobile ) ) {
```
premanshup: @UttamSharma007 Please use strict type checking.

inc/extras.php
UttamSharma007: Updated!

inc/extras.php
UttamSharma007: We can't use strict check because in some cases it may fail. The variable type can be boolean, string or int.

premanshup: Maybe we can do typecasting and check? As per the discussion, it can affect performance though the impact of typecasting is very less. @Nikschavan Sir, what should be do in such cases?

Nikschavan: We can handle changes to strict type checks in #1383 where it is not easily possible to implement.

PR #1451, opened 2019-09-10
fix: Menu open below header not working fix

premanshup:
Doc - https://wpastra.com/docs/display-sub-menu-below-the-header/


premanshup:
Merge this PR only after Batch Process PR #1240 is merged. Need to move this from theme update to batch processing function.

PR #1453, opened 2019-09-10
feat: Initial Site Title & Site Tagline partial refresh added.

premanshup:
…te Tagline.


inc/customizer/class-astra-customizer-partials.php
```diff
@@ -106,6 +106,20 @@ static function _render_header_main_rt_section_button_text() {
 			return do_shortcode( $custom_button_text );
 		}
 
+		/**
+		 * Render Partial Text Header Site Title & Tagline
+		 *
+		 * @since x.x.x
+		 */
+		static function _render_header_site_title_tagline() {
```
Nikschavan: @premanshup remove `_` from the function name. Also, does the method need to be static?

premanshup: @Nikschavan Sir, made changes:

1. Removed get_instance()
2. Removed `_` from the functions of Partial Refresh class.
3. Removed static ( working fine without static )
4. Footer bar partial refresh section 2 not working fix.

PR #1458, opened 2019-09-13
Style: dependency fix

patilvikasj:


inc/core/class-astra-enqueue-scripts.php
```diff
@@ -180,11 +180,20 @@ public function enqueue_scripts() {
 				// Register & Enqueue Styles.
 				foreach ( $styles as $key => $style ) {
 
+					$dependency = array();
+
+					// Add dynamic CSS dependency for all styles except for theme's style.css.
+					if ( 'astra-theme-css' !== $key && class_exists( 'Astra_Cache_Base' ) ) {
+						if ( ! Astra_Cache_Base::inline_assets() ) {
```
premanshup: I think we can use is_callable() function here?

patilvikasj: We have checked class_exists call already.

PR #1460, opened 2019-09-16
feat: Post format compatibility with excerpt.

patilvikasj:
…tent for video post format on blog page


inc/extras.php
```diff
@@ -1734,3 +1734,22 @@ function astra_wpforms_upgrade_link() {
 }
 
 add_filter( 'wpforms_upgrade_link', 'astra_wpforms_upgrade_link' );
+
+/**
+ * Decide whether to display except.
```
premanshup: except should be `excerpt`

inc/extras.php
```diff
@@ -1734,3 +1734,22 @@ function astra_wpforms_upgrade_link() {
...
+ *
+ * @since x.x.x
+ * @return bool
+ */
+function astra_display_excerpt() {
+
+	global $post;
+	$excerpt_type    = astra_get_option( 'blog-post-content' );
+	$display_excerpt = 'full-content' == $excerpt_type ? false : true;
```
premanshup: Strict typechecking

inc/extras.php
```diff
@@ -1813,3 +1813,22 @@ function astra_filter_socialsnap_upgrade_link( $link ) {
...
+
+/**
+ * Decide whether to display excerpt.
+ *
+ * @since x.x.x
+ * @return bool
+ */
+function astra_display_excerpt() {
+
+	global $post;
```
Nikschavan: This variable does not seem to be used anywhere, Can you verify and remove it?

inc/extras.php
```diff
@@ -1813,3 +1813,22 @@ function astra_filter_socialsnap_upgrade_link( $link ) {
...
+	$excerpt_type    = astra_get_option( 'blog-post-content' );
+	$display_excerpt = 'full-content' === $excerpt_type ? false : true;
+
+	$format = ( false !== get_post_format() ) ? get_post_format() : 'standard';
+
+	$display_excerpt = ( 'standard' !== $format ) ? false : $display_excerpt;
+
```
Nikschavan: The function right now has too many ternary conditions which is unnecessarily increasing the complexity of the function.

- We can skip returning boolean and instead return the `$excerpt_type` as 'full-content|excerpt' as the current option `blog-post-content` returns, This will reduce one condition.
- Rest of the conditions can be simipified by simply checking if post_format is not false and excerpt type is not 'full-content' if this conditon matches, force the excerpt to be `full-content` which is the condition that we want here.

```
function astra_excerpt_type() {
	$excerpt_type  = astra_get_option( 'blog-post-content' );

	if ( false !== get_post_format() && 'full-content' !== $excerpt_type ) {
		$excerpt_type = 'full-content';
	}

	return apply_filters( 'astra_excerpt_type', $excerpt_type );
}
```

Can you try if this will work as expected?

PR #1465, opened 2019-09-24
added CSS for color picker issue over 1920

UttamSharma007:


inc/customizer/custom-controls/color/color.scss
```diff
@@ -26,4 +26,10 @@
     width: 13px !important;
     margin-left: 8px !important;
   }
+}
+
+@media (min-width: 1900px) {
+  .iris-slider.iris-strip, .iris-strip.iris-slider.iris-alpha-slider {
+    margin-left: 8px !important;
```
premanshup: Can we remove `!important`?

UttamSharma007: No, !important is required!

Balachandark: The JS is added inline onto the element through JS so !important is necessary

PR #1466, opened 2019-09-24
added style and script html5 theme support

Balachandark:
* Added style and script html5 theme support

### DIFF
--- a/inc/class-astra-after-setup-theme.php
+++ b/inc/class-astra-after-setup-theme.php
@@ -107,6 +107,8 @@ if ( ! class_exists( 'Astra_After_Setup_Theme' ) ) {
 					'search-form',
 					'gallery',
 					'caption',
+					'style',
+					'script',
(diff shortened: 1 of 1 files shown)


inc/class-astra-after-setup-theme.php
```diff
@@ -107,6 +107,8 @@ public function setup_theme() {
 					'search-form',
 					'gallery',
 					'caption',
+					'style',
+					'script',
```
premanshup: Should we add this for WordPress version more than equal to 5.3? Will it throw any error for the older versions WordPress?

Balachandark: No, This won't throw any error as WordPress uses the theme support attributes through `current_theme_supports` function 
Like this: `current_theme_supports( 'custom-header', 'header-text' )`
and as they aren't using the style and script so there won't be any issue.

PR #1479, opened 2019-10-11
fix: Skip appending unit when value is inherit.

patilvikasj:


inc/core/common-functions.php
```diff
@@ -231,7 +231,7 @@ function astra_get_font_family( $value = '' ) {
 	 */
 	function astra_get_css_value( $value = '', $unit = 'px', $default = '' ) {
 
-		if ( '' == $value && '' == $default ) {
+		if ( ( '' == $value && '' == $default ) || ( 'inherit' === strtolower( $value ) || 'inherit' === strtolower( $default ) ) ) {
```
premanshup: @patilvikasj  Can we add strict type checking here?

patilvikasj: The value coming in a $value variable may come as string OR integer which can create a problem here.

PR #1501, opened 2019-11-04
revert: post format not standard changes

Balachandark:


inc/extras.php
```diff
@@ -1277,7 +1277,7 @@ function astra_the_excerpt() {
 
 		do_action( 'astra_the_excerpt_before', $excerpt_type );
 
-		if ( 'full-content' === astra_excerpt_type() ) {
+		if ( 'full-content' === $excerpt_type ) {
```
premanshup: @Balachandark We should provide a filter here for the users to show full content or excerpt.

Balachandark: added filter

PR #1523, opened 2019-11-19
Added author instead of author-name in creative work schema

GauravKhupse:
Fix: added author instead of author-name in creative work schema.


Nikschavan:
@GauravKhupse can you rebase this PR to the `next-release` branch?

PR #1533, opened 2019-12-03
improvement: Privilege to change default breakpoint of tablet & mobile device

Balachandark:


inc/core/common-functions.php
```diff
@@ -1316,3 +1316,48 @@ function astra_get_db_option( $option, $default = '', $deprecated = '' ) {
 	 */
 	return apply_filters( "astra_get_db_option_{$option}", $value, $option, $default );
 }
+
+/**
+ * Get the tablet breakpoint value.
+ *
```
premanshup: can we add since comment here?

inc/core/common-functions.php
```diff
@@ -1316,3 +1316,48 @@ function astra_get_db_option( $option, $default = '', $deprecated = '' ) {
...
+		$header_breakpoint = $header_breakpoint + $max;
+	}
+
+	return absint( $header_breakpoint );
+}
+
+/**
+ * Get the mobile breakpoint value.
+ *
+ * @param string $min min.
```
premanshup: can we add since comment here?

inc/core/common-functions.php
Balachandark: Since comment added

inc/compatibility/edd/class-astra-edd.php
```diff
@@ -390,6 +390,93 @@ public function add_styles( $assets ) {
 		 */
 		public function add_inline_scripts() {
 
+			$edd_css_output = '';
```
premanshup: @imnavanath Can we remove this and directly add it `$edd_css_output .= ` to `$edd_css_output = ` for the first occerence.

imnavanath: Corrected here - https://github.com/brainstormforce/astra/pull/1533/commits/22d60d967a10ea0fb498c92fe5aaea3ed932c647

PR #1540, opened 2019-12-06
feat: Global heading line height option

Balachandark:


inc/class-astra-dynamic-css.php
```diff
@@ -123,6 +128,9 @@ public static function return_output( $dynamic_css, $dynamic_css_filtered = '' )
 			if ( '' == $h1_text_transform ) {
 				$h1_text_transform = $headings_font_transform;
 			}
+			if ( '' == $h1_line_height ) {
```
premanshup: Why are we checking the condition here? It is always blank. 

Ref = https://github.com/brainstormforce/astra/pull/1540/files#diff-f250e1e66f981f4130ac58048586b6b1R117

`$h4_line_height = '';`

Please check the same for heading 4 to 6

inc/core/class-gutenberg-editor-css.php
```diff
@@ -85,6 +86,44 @@ public static function get_css() {
...
+			$h1_line_height = astra_get_option( 'line-height-h1' );
+			$h2_line_height = astra_get_option( 'line-height-h2' );
+			$h3_line_height = astra_get_option( 'line-height-h3' );
+
+			$h4_line_height = '';
+			$h5_line_height = '';
+			$h6_line_height = '';
+
+			// Fallback for H1 - headings typography.
+			if ( '' == $h1_line_height ) {
```
premanshup: Same as the above comment 

`if ( '' == $h1_line_height ) {` condition

inc/class-astra-dynamic-css.php
Balachandark: We get the option to check if it was set by the user before here - https://github.com/brainstormforce/astra/blob/44a3c6bd60b2993c0568dd66ff3a49cf79e6a68e/inc/class-astra-dynamic-css.php#L104

So if the value is already set in the database then global heading line-height would not be applied else it will be applied.

Balachandark: Changes done

inc/core/class-gutenberg-editor-css.php
Balachandark: Changes done

PR #1543, opened 2019-12-10
feat: Add fallback method batch migration

Balachandark:


inc/theme-update/class-astra-theme-background-updater.php
```diff
@@ -55,6 +55,64 @@ public function __construct() {
...
+						'sslverify' => apply_filters( 'https_local_ssl_verify', $sslverify ),
+					),
+				)
+			);
+
+			$cron_request['args']['blocking'] = true;
+
+			$result = wp_remote_post( $cron_request['url'], $cron_request['args'] );
+
+			if ( wp_remote_retrieve_response_code( $result ) >= 300 ) {
```
patilvikasj: Can we use is_wp_error instead?

patilvikasj: Cache failed result also in transient.

PR #1575, opened 2020-01-09
fix: Added HTML Tags support to Archive Description.

premanshup:


inc/core/common-functions.php
```diff
@@ -970,6 +970,16 @@ function astra_archive_page_info() {
...
 }
 
+/**
+ * Echo the archive description.
+ *
+ * Provided filter to add opening and closing tag for the archive description.
+ *
+ * @since x.x.x
+ */
+function astra_the_archive_description() {
```
Balachandark: Can we name this `astra_archive_description` instead of `astra_the_archive_description`?

premanshup: @Balachandark This function is no longer need, so I have removed this function. Please check the latest code.

PR #1577, opened 2020-01-15
fix: Button padding and border radius issue with Comment button and Search button.

premanshup:


premanshup:
We are closing this PR as the options from the customizer ( Border Radius, Padding, Line Height & Border Width  )should not work with Comment. 

This issue is fixed in the PR #1584

Issue #1585, opened 2020-01-24
Visual issues with LifterLMS certificates


thomasplevy:
Hello,

Today we've discovered a minor visual issue experienced by users using Astra with LifterLMS when they use certificates with a textured/colored background image.

This image depicts the issue:

![Screenshot_2020-01-24_14-39-59](https://user-images.githubusercontent.com/1290739/73109232-73258f00-3eb7-11ea-8ae6-2c0430c0b094.png)

The issue is the white box within the green and black lines. This box displays as a result of the following css rule: 

```css
.ast-separate-container .ast-article-post, .ast-separate-container .ast-article-single {
    background-color: #fff; /* this line causes the issue depicted above */
    border-bottom: 1px solid #eee;
    margin: 0;
    padding: 5.34em 6.67em;
}
```

Which I believe is generated dynamically based on rules added in the customizer.

A small update to the LifterLMS compatibility CSS file resolves the issue by explicitly declaring the background of this element as transparent.

```css
.llms-certificate-container .ast-article-single.llms_certificate {
    background-color: transparent;
}
```

I'll be sending along a pull request to resolve this issue shortly.

shubhw12:
@thomasplevy Thank you for bringing this to us. We had a look at this scenario in more detail. The solution provided by you works well but it will not be an ideal solution. Our team is working on this and it will be fixed in our future update. We really appreciate your concern about our product.

thomasplevy:
Of course, if you need any assistance from myself or the team please let me know.

shubhw12:
@thomasplevy We have fixed this issue and it will be released in our future updates. Thank you again for reporting this. here is the PR reference: #1594

thomasplevy:
@shubhw12 Thanks for the fix. I neglected to note a second way to create the issue in this issue report but I quietly addressed it in #1586 

I've made a note about it on #1594 

Thanks again <3

thomasplevy:
@shubhw12 we've been notified of what we think may be a regression which has caused this issue to recur on LifterLMS / Astra sites. The details have been outlined here: https://github.com/gocodebox/lifterlms/issues/1635

Outside of reproducing the issue I haven't had a developer working on investigating the source of this (whether it's a new conflict or a regression) but I wanted to let you know in case there's a quick fix from your side.

Thanks <3

PR #1586, opened 2020-01-24
fix: Define an explicit background color for element within LifterLMS certificate

thomasplevy:
Adds css to resolve issue described in #1585 

Additionally replaces text "WooCommerce" with "LifterLMS" in the file's file header comments.

### DIFF
--- a/assets/css/unminified/compatibility/lifterlms.css
+++ b/assets/css/unminified/compatibility/lifterlms.css
@@ -1156,6 +1156,7 @@ body .llms-loop-item-content .llms-progress-bar .progress-bar-complete {
   padding-left: 10px;
 }
 
-.llms-certificate-container .ast-article-single.llms_certificate {
+.llms-certificate-container .ast-article-single.llms_certificate,
+.llms-certificate-container .ast-article-single.llms_my_certificate {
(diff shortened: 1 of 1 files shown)


assets/css/unminified/compatibility/lifterlms.css
```diff
@@ -1155,3 +1155,8 @@ body .llms-loop-item-content .llms-progress-bar .progress-bar-complete {
 .ast-woocommerce-cart-menu .main-header-bar .main-header-log-out {
   padding-left: 10px;
 }
+
+.llms-certificate-container .ast-article-single.llms_certificate,
+.llms-certificate-container .ast-article-single.llms_my_certificate {
+    background-color: transparent;
+}
```
premanshup: @thomasplevy Thank you for the PR, we have added a fix of this in PR #1594. 

I think we can skip this CSS from the PR and can consider the changes you have added in the comment :-)

Can you please make the accordingly :-)

Thank you for this PR.

thomasplevy: Yes, i agree, the fix in #1594 seems much better than hacking at the issue via CSS.

PR #1597, opened 2020-02-10
fix: Google Font loading issue with FS_MTHOD ftpext.

premanshup:


inc/customizer/class-astra-font-families.php
```diff
@@ -135,7 +135,7 @@ public static function get_google_fonts() {
 					return array();
 				}
 
-				$file_contants     = astra_filesystem()->get_contents( $google_fonts_file );
+				$file_contants     = file_get_contents( $google_fonts_file ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
```
Balachandark: Please update the variable name to `file_contents`

PR #1619, opened 2020-02-25
chore: Added dev provision to load individual addons script if SCRIPT_DEBUG

imnavanath:
- Loading individual addon's scripts while SCRIPT_DEBUG is True


premanshup:
This PR is reverted in the #1625

PR #1623, opened 2020-03-02
fix: class for Availability text

pippyn:
If product was not in stock but on backorder. The class of the availability text would still be "stock in-stock". But it should be "stock available-on-backorder".


pippyn:
When do you plan on reviewing this PR? now I have to edit the file on all my sites after every update. It's just a simple fix in my opinion. Thanks

premanshup:
@pippyn Sorry for the delay, thank you for the PR :-)

If tested positive, this will be added after v2.4.4 update release.

premanshup:
@pippyn Thank you for this PR. This is merged for the next release.

PR #1653, opened 2020-03-13
 fix: bolt icon issue on Astra options page.

shubhw12:


functions.php
```diff
@@ -15,7 +15,7 @@
 /**
  * Define Constants
  */
-define( 'ASTRA_THEME_VERSION', '2.3.3' );
+define( 'ASTRA_THEME_VERSION', '2.3.3.1' );
```
Balachandark: Updated version should be 2.3.4

PR #1678, opened 2020-03-30
fix: RTL condition support for media query improvement

imnavanath:
- Adding RTL support for media query dynamic CSS


inc/class-astra-dynamic-css.php
```diff
@@ -998,22 +1075,21 @@ public static function return_output( $dynamic_css, $dynamic_css_filtered = '' )
 				'.ast-separate-container #secondary' => array(
 					'padding-top' => 0,
 				),
+				'.ast-separate-container.ast-two-container #secondary .widget' => array(
```
premanshup, line 1078: @imnavanath If this is related to RTL then why have we added this CSS?

imnavanath: This is not exactly for RTL, but we faced the spacing issue in widgets (in comparison with the master) while testing phase. So to resolve this we've added this CSS into dynamic. 

Here is the link for that - https://github.com/brainstormforce/astra/pull/1678/files#diff-57378523ad2be61fc9a2e391f07b2497L215

The last SCSS file in this PR

PR #1681, opened 2020-03-31
fix: Security nonce fix

imnavanath:
- Nonce created and verification of nonce added for extended plugins manager
- manage_options checked for security concern


inc/lib/notices/class-astra-notices.php
```diff
@@ -142,7 +142,13 @@ public function dismiss_notice() {
 		 * @return void
 		 */
 		public function enqueue_scripts() {
+
+			if ( ! current_user_can( 'manage_options' ) ) {
+				return;
+			}
```
premanshup, line 148: @imnavanath Please check if this will work for other plugin notices or not. What privilege does their plugin require? Is it always the same? Or maybe we can skip this? Let me know your thoughts.

imnavanath: Updated code with new filter.

PR #1707, opened 2020-04-15
improvement: BSF Analytics notice design improvements

imnavanath:
- Notice redesigned


admin/bsf-analytics/class-bsf-analytics.php
```diff
@@ -47,6 +54,24 @@ public function __construct() {
...
+		/**
+		 * BSF Analytics URL
+		 *
+		 * @return String URL of bsf-core directory.
+		 */
+		public function bsf_analytics_url() {
+
+			$path       = wp_normalize_path( BSF_ANALYTICS_PATH );
+			$theme_dir  = wp_normalize_path( get_template_directory() );
+			$plugin_dir = wp_normalize_path( WP_PLUGIN_DIR );
```
premanshup, line 66: @imnavanath Can we remove this Line? 

`$plugin_dir = wp_normalize_path( WP_PLUGIN_DIR );`

admin/bsf-analytics/class-bsf-analytics.php
```diff
@@ -47,6 +54,24 @@ public function __construct() {
 			$this->includes();
 		}
 
+		/**
+		 * BSF Analytics URL
+		 *
```
premanshup, line 59: @imnavanath Since comment is missing from the function comment.

imnavanath: @premanshup I've added the since comment (x.x.x) in the BSF-Analytics. Have a look - https://git.brainstormforce.com/wp-sharks/bsf-analytics/-/merge_requests/11/diffs

admin/bsf-analytics/class-bsf-analytics.php
imnavanath: Can't remove @premanshup, because we will load BSF-Analytics from other paid plugins as well so we can't restrict this condition for the theme only. We need to check for plugins as well.

premanshup: @imnavanath But we are not using `$plugin_dir` variable anywhere inside the function, so I'm not sure how it will be required?

imnavanath: Ohh yes, in that way. 
Got it, got it. Actually we removed the elseif condition, where it was used. But forget to remove this var.
Removed this now 👍

admin/bsf-analytics/class-bsf-analytics.php
imnavanath: Updated now 👍

PR #1719, opened 2020-04-22
improvement: Change method of loading Google Fonts from JSON to PHP Array.

premanshup:


inc/customizer/class-astra-font-families.php
```diff
@@ -129,16 +129,25 @@ public static function get_google_fonts() {
...
 
-				if ( ! file_exists( ASTRA_THEME_DIR . 'assets/fonts/google-fonts.json' ) ) {
+				if ( ! file_exists( $google_fonts_file ) ) {
 					return array();
 				}
 
-				$file_contants     = astra_filesystem()->get_contents( $google_fonts_file );
-				$google_fonts_json = json_decode( $file_contants, 1 );
+				if ( ( ASTRA_THEME_DIR . 'assets/fonts/google-fonts.json' ) === $google_fonts_json_file ) {
+					require_once $google_fonts_file;
```
imnavanath, line 140: Can we use directly `include` instead of `require_once`. So there will be no need for extra function declaration.

inc/customizer/class-astra-font-families.php
```diff
@@ -129,16 +129,25 @@ public static function get_google_fonts() {
...
+				$google_fonts_file      = apply_filters( 'astra_google_fonts_php_file', ASTRA_THEME_DIR . 'inc/google-fonts.php' );
 
-				if ( ! file_exists( ASTRA_THEME_DIR . 'assets/fonts/google-fonts.json' ) ) {
+				if ( ! file_exists( $google_fonts_file ) ) {
 					return array();
 				}
 
-				$file_contants     = astra_filesystem()->get_contents( $google_fonts_file );
-				$google_fonts_json = json_decode( $file_contants, 1 );
+				if ( ( ASTRA_THEME_DIR . 'assets/fonts/google-fonts.json' ) === $google_fonts_json_file ) {
```
imnavanath, line 139: Can you please once verify this condition @premanshup as we discussed.

inc/customizer/class-astra-font-families.php
premanshup: @imnavanath Thank you for the suggestion, done 👍

inc/customizer/class-astra-font-families.php
premanshup: @imnavanath Verified, this is correct. When the path is not default, load the json file else load the PHP array.

PR #1750, opened 2020-06-01
chore: v2.4.5 update.

premanshup:


inc/extras.php
```diff
@@ -1930,12 +1930,14 @@ function remove_controls( $wp_customize ) {
  *
  * @since 2.4.3
  * @param array $default_stats Default stats array.
- * @return array $default_stats Default stats with Theme specific stats array.
+ * @return array $default_stats Default stats with Theme specific stats array. Test Comment.
```
imnavanath, line 1933: Test Comment. should be removed now.

PR #1762, opened 2020-06-16
fix: Issues from QA in css-fixes-for-megamenu

imnavanath:
### Description
- Caret not visible for submenu in normal menu-submenu structure

### Screenshots
- Issue - https://share.getcloudapp.com/NQugDBOD
- With fix - https://share.getcloudapp.com/DOuAQJOG

### Types of changes
- Removed unwanted .sub-arrow selector from CSS & replaced with .menu-link
- Same is in master with anchor tag (so nothing will break - https://share.getcloudapp.com/12uNybxn)

### How has this been tested?
- With normal menu -submenu structure
- With Mega Menu
- With responsive device

### Checklist:
- [x] My code is tested


assets/css/unminified/style-rtl.css
```diff
@@ -3235,6 +3235,10 @@ body {
   align-self: center;
 }
 
+.ast-no-menu-items .main-navigation {
+  display: none;
+}
```
imnavanath, line 3240: Added this hiding CSS for if the menu is blank (when there is no men-item in it & last menu item is also none then it class appends to the header.)
Issue - Need to hide this empty nav space for above case - https://share.getcloudapp.com/wbuW7p6b
With the fix - https://share.getcloudapp.com/lluJYpo1

PR #1802, opened 2020-07-20
feat: Customizer Refresh Control PHP to ReactJS.

RajkiranBagal:
### Description
 - Convert Customizer Refresh from PHP to ReactJS.
 - Trello Task - https://trello.com/c/qlEbuDEW/21-ast-customizer-refresh

### Types of changes

### How has this been tested?
 - Check Backward Compatibility.
 - Check if all options are working and saving to database.

### Checklist:
- [x] My code is tested


premanshup:
@RajkiranBagal This PR should be opened for the Astra Pro plugin. Please verify once.

inc/customizer/react/src/customizer-refresh/control.js
```diff
@@ -0,0 +1,8 @@
+import CustomizerRefreshComponent from './customizer-refresh-component.js';
```
premanshup, line 1: @RajkiranBagal We should move this to the Astra Pro plugin. Or maybe we can keep it as it is as the HTML is not much and that too we are using this only once. Let me know your thoughts.

RajkiranBagal: @premanshup Sir
As to give this PR to Astra Pro will have to do all the setup for the React in Astra Pro.

So, I added that in the Astra theme. There is no issue for functionality.

Also, Yes there is not much HTML only one anchor tag, So we can keep it as it is.

But to maintain consistency I converted that to react.

premanshup: @RajkiranBagal As discussed, keeping this control in Addon only.

premanshup:
Closing this PR as this is added from Addon.

PR #1817, opened 2020-07-27
fix: added wp_print_styles for appearance submenu css.

shubhw12:
### Description
Instead of printing CSS directly on the admin-enqueue scripts, I have added this as an external CSS. This is done according to reference to a git hub issue 

### How has this been tested?
I have checked the places where this CSS applies.
I have checked the CSS in the source how it's been enqueued.
I have checked if this is loading on the front end.
I have checked this with RTL languages.

### Checklist:
- [ ] My code is tested


inc/core/class-astra-admin-settings.php
```diff
@@ -1649,6 +1642,28 @@ public static function top_header_right_section() {
 				<?php
 			}
 		}
+
+		/**
+		 * Add custom CSS for admin area sub menu icons.
+		 *
+		 * @since 2.6.0
```
premanshup, line 1649: @shubhw12 Since this function is added in the current version, it should be `x.x.x`

shubhw12: @premanshup Sir. I have changed it to x.x.x Please check once.

PR #1833, opened 2020-08-07
fix: removed extra padding applied for the block patterns in WP 5.5.

shubhw12:
### Description
Removed the extra padding applied for the block pattern in the latest WP5.5 update.

### Screenshots
Astra: https://share.getcloudapp.com/5zuGe7wL
Generate Press: https://share.getcloudapp.com/WnuJ1dld

### Types of changes
These changes shall not affect the front end but will fix the editor's view.

### How has this been tested?
checked this with the existing WP setup, also checked if CSS applied for the block pattern affect the main editor view.

### Checklist:
- [x] My code is tested


inc/core/class-gutenberg-editor-css.php
```diff
@@ -617,6 +617,38 @@ public static function get_css() {
...
+
+				$latest_editor_css = array(
+					'.ast-separate-container .block-editor-inserter__panel-content .block-editor-block-list__layout' => array(
+						'padding-top'    => '0px',
+						'padding-bottom' => '0px',
+						'padding-left'   => '0px',
+						'padding-right'  => '0px',
+
+					),
+					'.ast-page-builder-template .block-editor-block-list__layout' => array(
```
premanshup, line 631: @shubhw12 Please use some class to avoid this getting applied to the Page Editor as well, this should only be applied to Sidebar.

 https://a.cl.ly/eDuB9Zpp
https://a.cl.ly/8LuPweKZ

Please check this with all the CSS you have added.

inc/extras.php
```diff
@@ -313,6 +313,19 @@ function astra_attr( $context, $attributes = array(), $args = array() ) {
...
 
+	/**
+	 * Check the WordPress version.
+	 *
+	 * @since  x.x.x
+	 * @param string $version   WordPress version to compare with the current version.
+	 * @param string $compare   Comparison value i.e > or < etc.
+	 * @return bool            True/False based on the  $version and $compare value.
+	 */
+function astra_wp_version_compare( $version, $compare ) {
```
imnavanath, line 324: Instead of `astra_wp_version_compare` dedicated function to compare only WP version.
Can we create global `astra_version_compare` function which we can use it for any version_compare mechanism?

imnavanath: Ref - https://git.brainstormforce.com/astra/astra-addon/-/merge_requests/1716/diffs
Function - `astra_addon_do_version_cmpare()`

inc/core/class-gutenberg-editor-css.php
shubhw12: I have included the CSS conditionally as discussed before.

shubhw12: Removed the extra CSS as well

PR #1865, opened 2020-08-26
fix: removed the description links from customizer in case of white label

shubhw12:
### Description
Checked condition for the branding name for removing the description links from the customizer.

### How has this been tested?
checked with a white label on and off for theme in customizer.

### Checklist:
- [x] My code is tested


inc/customizer/configurations/class-astra-customizer-config-base.php
```diff
@@ -57,7 +57,7 @@ public function section_get_description( $args ) {
 
 			// Return if white labeled.
 			if ( class_exists( 'Astra_Ext_White_Label_Markup' ) ) {
```
imnavanath, line 59: @shubhw12 no need for this claas_exists check (in the above case as well 👆 ), in the function `astra_is_white_labelled` we already verified it. You can directly check `if ` with this function.

shubhw12: @imnavanath Valid.
I have removed the class exists condition check.

premanshup (approved):

imnavanath (approved):

inc/customizer/configurations/class-astra-customizer-config-base.php
imnavanath: 👍

PR #1977, opened 2020-10-03
perf: Use customize_controls_enqueue_scripts action for enqueueing Customizer scripts

westonruter:
### Description

I noticed when widgets are not included in the Customizer (such as when Astra is being used as a Reader theme in the AMP plugin), certain controls fail to get initialized.

The reason for this is that the wrong action is being used to add scripts in the Customizer. Instead of using `admin_enqueue_scripts` the `customize_controls_enqueue_scripts` script should be used. This ensures the script ias added only to the Customizer, and not to ever admin page.

The reason why it works most of the time (when Widgets are enabled) is that the Widgets component in the customizer does `do_action( 'admin_enqueue_scripts', 'widgets.php' )` for compatibility with the widgets admin screen.

### Screenshots

Before | After
-------|------
image | image

### Types of changes

Bugfix

### How has this been tested?

I created a child theme of Astra and added `AMP: true` to the `style.css`. Then I selected it as the Reader theme when selecting the Reader mode in the AMP plugin.

### Checklist:
- [ ] My code is tested


shubhw12:
Hello @westonruter 

Greetings for the day and thanks for the PR.

I tried to reproduce the issue at our end, I used the AMP plugin and set the reader theme, But I was not able to reproduce the scenario.

I had a look at the before-after screenshot that you have added, There I can see that the styles are quite messed up, we had the same scenario with the Gutenberg plugin ( from v8.9.0 to v9.1.0  ). The issue was fixed in one of the recent versions of the Gutenberg plugin ( v9.1.0 ).

Could you please if this same case at your end once.

westonruter:
Even if you can't reproduce the issue, the use of the `admin_enqueue_scripts` is incorrect because it means the scripts are getting enqueued on _every_ admin page. The scripts should only be enqueued in the Customizer, and to do that you should use the `customize_controls_enqueue_scripts` action.

premanshup (approved):

imnavanath (approved):

shubhw12 (approved):

PR #2083, opened 2020-12-07
Update class-astra-meta-boxes.php

Farhan25:
optimize $meta_options for loop for only its parameters not the whole $stored parameters

### Description
I changed the loop iteration from $stored to $meta_options as a reference using ampersand character.

### Screenshots
I tested it and it works fine, you may test it as well.

### Types of changes
only an optimization


vrundakansara:
@Farhan25 - Thank you for the PR, we will further test this.

premanshup:
@vrundakansara Can you please check if we have any update on this?

imnavanath (approved):

PR #2129, opened 2020-12-16
fix: added condition for pagination markup to avoid blank print.

sagarb90:
### Description
- Added condition for pagination markup to avoid blank print.

### Types of changes
- Improvement.

### Checklist:
- [x] My code is tested


inc/markup-extras.php
```diff
@@ -141,10 +141,14 @@ function astra_body_classes( $classes ) {
 	 * @return void            Generate & echo pagination markup.
 	 */
 	function astra_number_pagination() {
-		global $numpages;
+		global $wp_query;
 		$enabled = apply_filters( 'astra_pagination_enabled', true );
 
-		if ( isset( $numpages ) && $enabled ) {
+		// Don't print empty markup if their is only one page.
+		if ( ( $wp_query->max_num_pages < 2 && ( is_home() || is_archive() || is_search() ) ) || ! $enabled ) {
```
premanshup, line 148: @sagarb90 Any specific reason for adding this condition?

`( is_home() || is_archive() || is_search() ) )`

premanshup (changes requested):

inc/markup-extras.php
sagarb90: @premanshup Yes pagination markup is added in these 3 pages so added conditions for the same.

inc/markup-extras.php
```diff
@@ -141,10 +141,14 @@ function astra_body_classes( $classes ) {
 	 * @return void            Generate & echo pagination markup.
 	 */
 	function astra_number_pagination() {
-		global $numpages;
+		global $numpages, $wp_query;
 		$enabled = apply_filters( 'astra_pagination_enabled', true );
 
-		if ( isset( $numpages ) && $enabled ) {
+		// Don't print empty markup if their is only one page.
+		if ( ( isset( $numpages ) && $wp_query->max_num_pages < 2 ) || ! $enabled ) {
```
premanshup, line 148: @sagarb90 Can we use $numpages instead of `$wp_query->max_num_pages` here? This way maybe we can remove `$wp_query;`?

premanshup (changes requested):

inc/markup-extras.php
sagarb90: @premanshup No we cannot use $numpages as it always return 1 for single page with multiple posts, if no. of post more than 3 and page limit is 2 then we are getting $numpages value =1 so we get the exact count in $wp_query->max_num_pages only, so I have used this I have also refer other themes pagination condition they have use same for that.

premanshup (approved):

PR #2152, opened 2020-12-21
improvement: revamp main header section div structure

sagarb90:
### Description
- Revamp Desktop Header div structure. 
- Reduce no. of divs.
 
### Types of changes
- Improvements.
 
### Checklist:
- [x] My code is tested


template-parts/header/builder/header-row.php
```diff
@@ -20,75 +20,72 @@
...
+								do_action( 'astra_render_header_column', $row, 'left' );
+							if ( Astra_Builder_Helper::has_center_column( $row ) ) {
+								?>
+										<div class="site-header-<?php echo esc_attr( $row ); ?>-section-left-center site-header-section ast-flex ast-grid-left-center-section">
 									<?php
 									/**
 									 * Astra Render Header Column
 									 */
-									do_action( 'astra_render_header_column', $row, 'center' );
+									do_action( 'astra_render_header_column', $row, 'left_center' );
```
premanshup, line 49: @sagarb90 Please verify this in theme and addon if we are changing the action it should be changed from everywhere else too.

premanshup (changes requested):

template-parts/header/builder/header-row.php
sagarb90: @premanshup We have not changing any action here we have just work on Div structure I think  due to line number change it showing that changes I have also compare this with main branch file so none of action is missing or changed.

premanshup (approved):

vrundakansara (approved):

imnavanath (approved):

PR #2157, opened 2020-12-22
Improvements: Removed site container div and modify  it's depended CSS.

sagarb90:
### Description
- Removed site container div and modify  it's depended CSS.
- Removed from Desktop header, mobile header and footer layout.

### Types of changes
- Improvements.

### Checklist:
- [x] My code is tested


template-parts/header/builder/mobile-header-row.php
```diff
@@ -37,7 +37,6 @@
 				 */
 				do_action( "astra_header_{$row}_container_before" );
 				?>
-				<div class="site-container">
```
premanshup, line 40: @sagarb90 Please check if there is any CSS related to in Addon as well? We should remove it as well.

premanshup (changes requested):

template-parts/header/builder/mobile-header-row.php
sagarb90: @premanshup I have checked their is no any CSS or class related to  `site-container` in add-on.

vrundakansara (approved):

PR #2258, opened 2021-01-22
improvement: Revamp CSS and DIV structure of Comment module

sagarb90:
### Description
- Revamp CSS and DIV structure of Comment module.
- Replace float property with flex.
- Removed Extra divs and manage it's CSS.
- Also removed `clear: both` property CSS as we using flex property so no need this.

### Types of changes
- Improvements

### How has this been tested?
- Test all type of comments design and check if it is breaking layout.

### Checklist:
- [x] My code is tested


premanshup:
@sagarb90 Just wondering why there are no changes in the Static files? Can you please once verify if any changes in the SCSS files for comment required?

sagarb90:
@premanshup As we have move all the static CSS to dynamic CSS  for the comment module so their is no change in SCSS file.

premanshup:
> @premanshup As we have move all the static CSS to dynamic CSS for the comment module so their is no change in SCSS file.

@sagarb90 Ah right! The PR is looking good 👍🏼

comments.php
```diff
@@ -30,7 +30,6 @@
 	<?php astra_comments_before(); ?>
 
 	<?php if ( have_comments() ) : ?>
-		<div class="comments-count-wrapper">
```
premanshup, line 33: @sagarb90 Need to manage backward compatibility for HTML markup with old structure.

premanshup (changes requested):

comments.php
sagarb90: Added  backward compatibility using common markup function.

premanshup (changes requested):
@sagarb90 Instead of overriding the CSS, can we have just one CSS selector? Maybe we can add layout specific CSS Classes and add them conditionally.

For eg. - [https://d.pr/i/FdDAWj](https://d.pr/i/FdDAWj)

sagarb90 (commented):
`Instead of overriding the CSS, can we have just one CSS selector? Maybe we can add layout specific CSS Classes and add them conditionally` - This is Done, Added layout condition for border-bottom CSS.

premanshup (approved):

PR #2306, opened 2021-02-04
feat: Added units to slider controls

abhishek-bsf:
### Description

Added units to slider controls
Fixed display issue with slider units

### Types of changes

### How has this been tested?

Tested if Units are displayed after all slider controls

### Checklist:
- [x] My code is tested


inc/customizer/custom-controls/assets/css/unminified/slider.css
```diff
@@ -0,0 +1,12 @@
+.customize-control-ast-slider .wrapper {
+  display: flex;
+}
+
+.customize-control-ast-slider .components-range-control {
+  width: -webkit-fill-available;
```
imnavanath, line 6: webkit will only work on chrome, what about other browsers? What's the case here? I can help.

imnavanath: Used 100% width

PR #2338, opened 2021-02-09
fix: white label not working in copyright component

shubhw12:
### Description
Instead of using static values for theme author and URL, I have used the variables just like what was done previously previous header footer markup.

### Types of changes

Bug fix (non-breaking change which fixes an issue)

### How has this been tested?
Checked this with the white label enabled.

### Checklist:
- [x] My code is tested


inc/builder/markup/class-astra-builder-footer.php
```diff
@@ -223,13 +223,21 @@ public function render_column( $row, $column ) {
 		 */
 		public function footer_copyright() {
 
+			$theme_author = apply_filters(
+				'astra_theme_author',
+				array(
+					'theme_name'       => __( 'Astra WordPress Theme', 'astra' ),
+					'theme_author_url' => 'https://wpastra.com/',
+				)
+			);
```
premanshup, line 232: @shubhw12 Instead of adding two apply filter for same filter, it will be recommended to create a single function with single apply_filter and call that function here.

premanshup (changes requested):

Nikschavan (approved):

inc/builder/markup/class-astra-builder-footer.php
shubhw12: @premanshup Sir I have made the changes accordingly

vrundakansara (approved):

premanshup (approved):

PR #2387, opened 2021-02-19
fix: Label improvements

vrundakansara:
### Description
Label improvements

### Checklist:
- [x] My code is tested


inc/customizer/extend-custom-controls/src/social-icons/item-component.js
```diff
@@ -18,6 +18,7 @@ const ItemComponent = props => {
 	});
 	
 	let icon = props.item.id.replace(/[\d_]+$/g, '');
+	let urlLabel = ( 'phone' === props.item.id ) ? __('Number', 'astra') : __('URL', 'astra');
```
premanshup, line 21: We can use const here.

vrundakansara: Done, please re verify.

rohitp-bsf (approved):

premanshup (approved):

inc/customizer/extend-custom-controls/src/social-icons/item-component.js
```diff
@@ -17,7 +17,8 @@ const ItemComponent = props => {
 		open: false,
 	});
 	
-	let icon = props.item.id.replace(/[\d_]+$/g, '');
+	const icon = props.item.id.replace(/[\d_]+$/g, '');
```
Nikschavan, line 20: Usually with Regex, I prefer writing a description inline as to what this is catching - Because if in the future a Regex is to be modified, it is important to know all past cases why it was written etc.

vrundakansara: @Nikschavan - As of now, this PR doesn't really add the Regex here. This PR addresses the `let`/`const` change. Perhaps I can add this as feedback and get the purpose of the Regex added in the comment later in [this task](https://trello.com/c/q6LHjtAi/266-feedback-from-nikhil-on-the-prs-common-function).

PR #2407, opened 2021-02-24
 Add web stories embed support with default customizer settings

pradeep910:
### Description
Adds web stories circles list above the header if web stories are enabled via customizer settings. This uses web stories added to the site using [Web Stories Plugin](https://wordpress.org/plugins/web-stories/). 

Note: In order to make it work please use this [web-stories-v7.zip](https://github.com/brainstormforce/astra/files/6035970/web-stories-v7.zip) as this feature is not yet released. Here is the [github branch link](https://github.com/rtCamp/web-stories-wp/tree/feature/embed-stories).

### Screenshots
Screenshot 2021-02-24 at 5 31 06 PM
Screenshot 2021-02-24 at 5 31 38 PM

### Types of changes
- Adds theme customizer settings support using `add_theme_support( 'web-stories' );`
- Changes `header.php` to add web stories embed code. (This displays web stories above the header on all pages.)

### How has this been tested?

Tested with a few astra starter templates (Photographer, Custom Printing, Italian Restaurant) with customizer enabled stories in desktop, tablet and mobile devices.
(Please use this [web-stories-v7.zip](https://github.com/brainstormforce/astra/files/6035970/web-stories-v7.zip) as this feature is not yet released. Also, this will require "Show stories" to be enabled from customizer.)

### Checklist:
- [x] My code is tested


pradeep910:
@Nikschavan How would you like this PHP 5.2 issue resolved? Looks like astra requires PHP 5.3, so we may not need to resolve this. https://github.com/brainstormforce/astra/blob/master/style.css#L10

```
FILE: /home/runner/work/astra/astra/header.php
----------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 1 LINE
----------------------------------------------------------------------
 35 | ERROR | the \ operator (for namespaces) is not present in PHP
    |       | version 5.2 or earlier
    |       | (PHPCompatibility.LanguageConstructs.NewLanguageConstructs.t_ns_separatorFound)
 35 | ERROR | the \ operator (for namespaces) is not present in PHP
    |       | version 5.2 or earlier
    |       | (PHPCompatibility.LanguageConstructs.NewLanguageConstructs.t_ns_separatorFound)
 35 | ERROR | the \ operator (for namespaces) is not present in PHP
    |       | version 5.2 or earlier
    |       | (PHPCompatibility.LanguageConstructs.NewLanguageConstructs.t_ns_separatorFound)
----------------------------------------------------------------------
```

inc/class-astra-after-setup-theme.php
```diff
@@ -175,6 +175,9 @@ public function setup_theme() {
 				);
 			}
 
+			// Enable Webstories.
+			add_theme_support( 'web-stories' );
```
Nikschavan, line 179: This can be moved in the above-mentioned file as well, the idea is - everything related to Web stories will be initiated from this file.

sass/site/_site.scss
```diff
@@ -47,3 +47,6 @@
 @import "compatibility/page-builder/page-builder";
 
 @import "layout/flex";
+
+/*----------  Web Stories  ----------*/
+@import "web-stories/web-stories-circles";
```
Nikschavan, line 52: Ideally, we would want to load the CSS in the frontend only if the web-stories plugin is active and a dtory it is being displayed in the page.

If the CSS is going to be small like it is right now this can be added in the dynamic CSS (CSS generated using PHP, where we can use conditionals) and an example for this is for [elementor comparibility](https://github.com/brainstormforce/astra/blob/master/inc/compatibility/class-astra-elementor.php#L57), we CSS in the form of an array to the filter `astra_dynamic_theme_css`.

Since we don't have public documentation for this filter, feel free to ignore this change, I can get this done later on.

pradeep910:
@Nikschavan I see there are some namespace related issues coming from PHP 5.2. WordPress needs minimum 5.6, so may I suggest to update php version compatibility check?

Also please note WP also needs minimum 5.6 version to run. https://wordpress.org/about/requirements/

sass/site/web-stories/_web-stories-circles.scss
```diff
@@ -0,0 +1,4 @@
+.web-stories-list.web-stories-list--customizer.is-view-type-circles {
+	border-bottom: 1px #ccc solid;
+	padding: 15px 0;
```
Nikschavan, line 3: Is there going to be more CSS than this to add compatibility? 
Is this padding going to be required by all the themes? If this is the case I was wondering if this can be moved into the plugin? That would reduce the effort required from the theme authors to add compatibility for their themes.

Nikschavan (changes requested):
Adding some feedback on the overall structure we are following to add compatibility to third-party plugins.

Nikschavan:
@pradeep910 Re: PHP 5.2, You can ignore this rule in the new file that will be created, This is done for the elementor compatibility file I mentioned above. We are planning to raise the minimum PHP version but we will have to do it with an announcement to the users and that has just been pending to be done. 

While WP raised the minimum version, themes have to support the last three versions of WordPress so we could raise the minimum version since WP 5.5. Although, we appreciate that there might be sites among 1M active installs that might be updating theme but not WordPress so we feel it is necessary to announce it before upping the requirements.

---

btw - I tried web stories on a few starter templates - I noticed that if the user has not used "Global background color" (Global -> Colors -> Base Colors) then the stories container does not have any background color - 

Markup 2021-02-24 at 21 10 31

Do you think it makes sense for the web-stories plugin to add some basic styling options?

The global background color will not really work for this case as this color is added everywhere on the site where the background color is not added to any div. For eg try removing this color from the site that you are using in the screenshot.

I believe a theme would need to provide separate options anyway to style and make this section look nicer and integrated in the overall UI. Let me know your thoughts on this.

sass/site/_site.scss
Nikschavan: Also - This can be decided after the whole compatibility CSS will be ready.

header.php
```diff
@@ -28,6 +28,14 @@
 </head>
 
 <body <?php astra_schema_body(); ?> <?php body_class(); ?>>
+<?php
+
+// Embed web stories above header with pre-configured customizer settings.
+if ( function_exists( '\Google\Web_Stories\render_theme_stories' ) ) {
+	\Google\Web_Stories\render_theme_stories();
+}
```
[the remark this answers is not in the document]
pradeep910: Creating separate file as you mentioned makes sense, however we wanted to keep changes minimum to demonstrate how stories can be shown above header.

There is another hook on top of `wp_body_open`, which is Astra's own hook - `astra_body_top`. As that can take precedence over `wp_body_open` we can try that maybe?

sass/site/_site.scss
pradeep910: I liked the idea of adding css dynamically. It totally depends on theme author how much they would like to support and how they would like to style stories on a particular template. 

In general, for a simple template or design it may not make sense to add css but for complex design where you can have sticky elements and absolute positioned elements it may need extra css to override. 

We will have to determine how much css changes are needed in the existing free starter templates https://wordpress.org/plugins/astra-sites/ and the premium starter templates https://wpastra.com/docs/install-premium-starter-templates/

pradeep910:
Thanks for the feedback about global color @Nikschavan. 

@divyarajmasani do you have any suggestion on this?

sass/site/_site.scss
Nikschavan: Agreed - The CSS might be bit more depending on all the options that are possible.

sass/site/web-stories/_web-stories-circles.scss
pradeep910: As I mentioned in my [previous comment](https://github.com/brainstormforce/astra/pull/2407#discussion_r582089646), it might need more css depending upon the starter template used. Some templates might need extra CSS, but adding them dynamically will be a good idea. We can create array of templates used if Astra stores currently used template somewhere and then add CSS based on that. 
Not sure if it is too much to check just to add one line of css.

divyarajmasani:
@Nikschavan 

That would be a really nice addition to the web-stories plugin, users being able to tweak some basic styles. We would keep this in our future roadmap and definitely try to address it as quickly as possible. 

Thank you for your feedback.

inc/compatibility/class-astra-web-stories.php
```diff
@@ -0,0 +1,89 @@
...
+
+		/**
+		 * Initiator
+		 */
+		public static function get_instance() {
+			if ( ! isset( self::$instance ) ) {
+				self::$instance = new self();
+			}
+			return self::$instance;
+		}
```
premanshup, line 44: Just a suggestion, instead of the singleton class can we keep it as a regular class?
```suggestion

```

inc/compatibility/class-astra-web-stories.php
```diff
@@ -0,0 +1,89 @@
...
+		} // end function web_stories_embed
+
+	}
+
+endif;
+
+/**
+ * Kicking this off by calling 'get_instance()' method
+ */
+Astra_Web_Stories::get_instance();
```
premanshup, line 89:
```suggestion
new Astra_Web_Stories();
```

inc/compatibility/class-astra-web-stories.php
```diff
@@ -0,0 +1,89 @@
...
+		/**
+		 * Custom render function for Web Stories Embedding.
+		 */
+		public function web_stories_embed() {
+			// Embed web stories above header with pre-configured customizer settings.
+            if ( function_exists( '\Google\Web_Stories\render_theme_stories' ) ) {
+                \Google\Web_Stories\render_theme_stories();
+
+				?>
+				<style type="text/css">
```
pradeep910, line 71: @premanshup @Nikschavan I think this CSS can be added through `astra_dynamic_theme_css` filter available in theme. You can take over and change it to as per your structure or standards.

Nikschavan:
@pradeep910 Can you check if in this PR "Allow edits by maintainers" setting is enabled?

Markup 2021-03-11 at 18 31 01

With this we can add final changes to this PR and get this merged, Right now I don;t have permission to add any commits to this PR.

Nikschavan:
> @pradeep910 Can you check if in this PR "Allow edits by maintainers" setting is enabled?
> 
> Markup 2021-03-11 at 18 31 01
> 
> With this we can add final changes to this PR and get this merged, Right now I don;t have permission to add any commits to this PR.

@pradeep910, @divyarajmasani - Can you check this setting? We are good to merge this after fixing the phpcs waning and moving the CSS to use `astra_dynamic_theme_css` filter. I can get this done and get the PR merged.

pradeep910:
I don't see this option somehow. @Divyaraj Masani
***@***.***> please check.

Regards,
Pradeep

Nikschavan:
That is weird, I am not able to push to this PR as I don't have that permission (this option allows maintainers of the repo to push commits to the PRs)

I have created a separate PR from this one #2488 and will be getting that one merged soon, You can add any feedback if you have there.

PR #2454, opened 2021-03-06
refactor: Added 2px border support for outline GB buttons in all cases

imnavanath:
### Description
- As per requirement 2px border will apply to all outline GB buttons

### Types of changes
- Dynamic CSS changes for outline buttons adjusting border size

### How has this been tested?
- Test this with Customizer > Global > Buttons - Border width
- Check with empty border size (in this case border should be 2px for outline GB buttons)
- Check with '0' border size (in this case border should be 2px for outline GB buttons)
- Check with 'x' number border size (in this case border should be 'x' px for outline GB buttons)

### Checklist:
- [x] My code is tested


assets/js/unminified/customizer-preview.js
```diff
@@ -1229,16 +1229,20 @@ function isJsonString( str ) {
...
-					dynamicStyle += 'border-top-width:'  + border.top + 'px;';
-					dynamicStyle += 'border-right-width:'  + border.right + 'px;';
-					dynamicStyle += 'border-left-width:'   + border.left + 'px;';
-					dynamicStyle += 'border-bottom-width:'   + border.bottom + 'px;';
-					dynamicStyle += 'border-style: solid;';
-					dynamicStyle += '}';
-
-				astra_add_dynamic_css( 'theme-button-border-group-border-size', dynamicStyle );
+				if( astraCustomizer.gb_outline_buttons_patterns_support ) {
+					wp.customize.preview.send( 'refresh' );
```
vrundakansara, line 1233: Do you think we should add some notes to the customizer? This is because the user will get confused about why his borders won't apply?
@premanshup - Your thoughts here?

vrundakansara (changes requested):

assets/js/unminified/customizer-preview.js
imnavanath: Borders will apply here, but in this pattern compatibility update we managed border & padding by some calculations, these will be bit tricky in preview JS that's why added here by refreshing preview.
@premanshup ☝️

premanshup: @vrundakansara as @imnavanath has mentioned that the borders will apply then we can skip the notice I think.

vrundakansara (approved):

rohitp-bsf (approved):

premanshup (approved):

PR #2474, opened 2021-03-09
refactor: Improved performance by using async functions with promises

rohitp-bsf:
### Description
- As we need to register settings and controls initially before using, it's usually taking a time to loop over thousands of array of the objects which blocking a customizer page to load.
- Made function asynchronous and wait for the promise to complete which will not wait for the entire loop and dependent function will execute after the async execution.

### Screenshots
NA

### Types of changes
- Made possible functions asynchronous.
- Added dependent functions inside promises. 

### How has this been tested?
- Check the performance benefit by comparing the previous version.
- Check possible actions like partial refresh, tab switch, grid drop, etc.

### Checklist:
- [ ] My code is tested


vrundakansara (approved):

inc/assets/js/customizer-preview.js
```diff
@@ -38,6 +38,22 @@
             }
         );
 
+		/**
+		 * Register partial refresh events at once asynchronously.
+		 */
+		wp.customize.preview.bind( 'active', function() {
+			var partials = $.extend({}, astraCustomizer.dynamic_partial_options);
+			Object.keys(partials).forEach(async function ( key) {
```
premanshup, line 46: @rohitp-bsf This is ES6 nad may not work in the browsers where ES6 is not supported. Here is a polyfill which can be used for this - 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#polyfill

package.json
```diff
@@ -32,9 +32,10 @@
     "kind-of": ">=6.0.3",
     "minimist": ">=1.2.3",
     "node-sass": "^4.14.1",
-    "postcss-flexibility": "^2.0.0"
+    "postcss-flexibility": "^2.0.0",
+    "@wordpress/scripts": "^12.6.1"
```
premanshup, line 36: @rohitp-bsf Can you please explain why we are including this in our main package.json file?

inc/assets/js/customizer-preview.js
rohitp-bsf: Resolved! @premanshup

package.json
rohitp-bsf: Removed! Thanks @premanshup

premanshup (approved):

PR #2486, opened 2021-03-11
refactor: Partial refresh calls to the backend.

rohitp-bsf:
### Description
- While doing partial refresh it was recalculation static array which is the same always. 

### Screenshots
- NA

### Types of changes
- Created a database option with autoloading false.
- While doing the partial refresh, fetching the database and reset the existing array instead of recalculating.
- Delete cached data on publishing.

### How has this been tested?
- Check the possibility of breaking the application.
- Check partial /. page refresh.

### Checklist:
- [ ] My code is tested


inc/customizer/class-astra-customizer.php
```diff
@@ -988,11 +1001,12 @@ public function starts_with( $string, $start_string ) {
 		/**
 		 * Filter and return Customizer Configurations.
 		 *
-		 * @param WP_Customize_Manager $wp_customize Reference to WP_Customize_Manager.
 		 * @since 1.4.3
 		 * @return Array Customizer Configurations for registering Sections/Panels/Controls.
 		 */
-		private function get_customizer_configurations( $wp_customize ) {
+		private function get_customizer_configurations() {
```
vrundakansara, line 1007: Any reason why we have removed the function parameter and added a global variable?

inc/customizer/class-astra-customizer.php
```diff
@@ -187,11 +186,19 @@ public function filter_dynamic_setting_args( $setting_args, $setting_id ) {
 			 * Prepare Contexts and choices.
 			 *
 			 * @since 3.0.0
-			 * @param object $wp_customize customizer object.
 			 */
-		public function prepare_customizer_javascript_configs( $wp_customize ) {
+		public function prepare_customizer_javascript_configs() {
```
vrundakansara, line 190: Any reason why we have removed the function parameter and added a global variable?

inc/customizer/class-astra-customizer.php
```diff
@@ -187,11 +186,19 @@ public function filter_dynamic_setting_args( $setting_args, $setting_id ) {
...
 
-			$configurations = $this->get_customizer_configurations( $wp_customize );
+			global $wp_customize;
+
+			$cached_data = get_option( 'astra_partials_config_cache', false );
+
+			if ( $wp_customize->selective_refresh->is_render_partials_request() && $cached_data ) {
+				self::$dynamic_options = $cached_data;
```
vrundakansara, line 197: Where have we assigned the `self::$dynamic_options` when it doesn't go in this `if`.
First-time user, a first-time refresh of the customizer, who has no cache saved?

vrundakansara (changes requested):

inc/customizer/class-astra-customizer.php
rohitp-bsf: `$wp_customize` variable is defined by WP and can be used as global anywhere. Passing it as the function will unnecessarily create extra injection to the function which may increase a `Cyclomatic complexity` in the future. @vrundakansara

inc/customizer/class-astra-customizer.php
rohitp-bsf: `$wp_customize` variable is defined by WP and can be used as global anywhere. Passing it as the function will unnecessarily create extra injection to the function which may increase a `Cyclomatic complexity` in the future. @vrundakansara

inc/customizer/class-astra-customizer.php
rohitp-bsf: As we have added settings and partials dynamically, we are having filters for that where we are identifying these settings and partials at the server end which is initially stored in `self::$dynamic_options` variable, and on each refresh, it is unnecessarily calculating again and again.
so,

1. initially we calculate everything as it was and store in the database only settings and partials.
2. on each request, check if is partial rendering? and have cached data already then return cached or again recalculate as a fallback.
3. delete the cached data on the customizer save. and repeat at step 2 as a fallback. @vrundakansara

vrundakansara: Okay.

vrundakansara (dismissed):

inc/customizer/class-astra-customizer.php
```diff
@@ -187,11 +186,19 @@ public function filter_dynamic_setting_args( $setting_args, $setting_id ) {
...
 			 * @since 3.0.0
-			 * @param object $wp_customize customizer object.
 			 */
-		public function prepare_customizer_javascript_configs( $wp_customize ) {
+		public function prepare_customizer_javascript_configs() {
 
-			$configurations = $this->get_customizer_configurations( $wp_customize );
+			global $wp_customize;
+
+			$cached_data = get_option( 'astra_partials_config_cache', false );
```
premanshup, line 194: @rohitp-bsf The PR is looking good but I'm skeptical about one thing. I see that we are reducing the iterations of the loop but we are also increasing a DB call on every load. Just confirming if this will not affect the performance? Can you please share stats of by how much this is increasing the performance? Thank you!

premanshup (changes requested):

inc/customizer/class-astra-customizer.php
rohitp-bsf: How it was working?

- While preparing the configs for each control config default we were using the `astra_get_option` which is implicitly calling the database around `no of configs` times.
- On each partial refresh again the same thing where we do not need this as we have migrated settings to the JS.
- We just need the partial array for the WP dynamic filters ex `customize_dynamic_partial_args`

How it is working now?
- We are preparing the partial array at once on the first customizer load.
- On each partial refresh we are hitting the database just for once.


Taking about the performance: I am trying to major the partial performance time though but I can see sufficient speed improvement in the partial refresh.

@premanshup

premanshup: @rohitp-bsf Got it! Thank you for the explanation 👍🏼

premanshup (approved):

premanshup (dismissed):

vrundakansara (dismissed):

premanshup (approved):

vrundakansara (approved):

PR #2596, opened 2021-04-07
fix: Added a type cast for boolean

rohitp-bsf:
### Description
- Fix: Added a safe case to typecast string to bool due to strict check.

### Checklist:
- [ ] My code is tested

### DIFF
--- a/inc/core/builder/class-astra-builder-helper.php
+++ b/inc/core/builder/class-astra-builder-helper.php
@@ -801,7 +801,7 @@ final class Astra_Builder_Helper {
 	public static function is_header_footer_builder_active() {
 
 		$astra_settings                             = get_option( ASTRA_THEME_SETTINGS );
-		$astra_settings['is-header-footer-builder'] = isset( $astra_settings['is-header-footer-builder'] ) ? $astra_settings['is-header-footer-builder'] : true;
+		$astra_settings['is-header-footer-builder'] = isset( $astra_settings['is-header-footer-builder'] ) ? (bool) $astra_settings['is-header-footer-builder'] : true;
 		return apply_filters( 'astra_is_header_footer_builder_active', $astra_settings['is-header-footer-builder'] );
 	}
 


inc/core/builder/class-astra-builder-helper.php
```diff
@@ -801,7 +801,7 @@ public function deprecate_old_header_and_footer( $args ) {
 	public static function is_header_footer_builder_active() {
 
 		$astra_settings                             = get_option( ASTRA_THEME_SETTINGS );
```
Nikschavan, line 803: Out of curiosity - why are we not using `astra_get_option` here?

vrundakansara: The action executes way before the defaults are set for `astra_get_option`.
That is why we have used native WP function here @Nikschavan

vrundakansara (approved):

Nikschavan (approved):

Nikschavan (approved):

inc/core/builder/class-astra-builder-helper.php
```diff
@@ -800,9 +800,7 @@ public function deprecate_old_header_and_footer( $args ) {
 	 */
 	public static function is_header_footer_builder_active() {
 
-		$astra_settings                             = get_option( ASTRA_THEME_SETTINGS );
-		$astra_settings['is-header-footer-builder'] = isset( $astra_settings['is-header-footer-builder'] ) ? $astra_settings['is-header-footer-builder'] : true;
-		return apply_filters( 'astra_is_header_footer_builder_active', $astra_settings['is-header-footer-builder'] );
+		return apply_filters( 'astra_is_header_footer_builder_active', (bool) astra_get_option( 'is-header-footer-builder', true ) );
```
vrundakansara, line 803: We will need the original `get_option` code @rohitp-bsf here, since this executes way before `astra_get_option` defaults are set.

vrundakansara (changes requested):

inc/core/builder/class-astra-builder-helper.php
rohitp-bsf: Right! Updated the code.

vrundakansara (approved):

inc/core/builder/class-astra-builder-helper.php
Nikschavan: Let's also add a comment here stating that, please. so tht we don't get this question again in future.

PR #2616, opened 2021-04-09
Add changes back which were removed accidently when merging master to next-release

Nikschavan:
### Description
In PR  #2614 some cade was removed accidentally, this PR adds that back to next-release branch.


assets/js/unminified/frontend.js
```diff
@@ -436,11 +436,6 @@ var astraTriggerEvent = function astraTriggerEvent( el, typeArg ) {
 
 		// Skip resize event when keyboard display event triggers on devices.
 		if( 'INPUT' !== document.activeElement.tagName ) {
-
-			if( menu_toggle_close ) {
-				menu_toggle_close.click();
-			}
-
```
Nikschavan, line 443: @RajkiranBagal can you confirm this change - that this needs to be removed from the `next-release` branch?

Right now in `next-release` this is throwing a JS error - 

Astra Workflow Next Release – Just another WordPress site 2021-04-09 at 1 53 38 PM

RajkiranBagal: This code is needed. Along with the Variable defination. The error is because the definition of variable seems missing.

Nikschavan: Can you push that to this branch?

RajkiranBagal: Yes.

RajkiranBagal: Actually, I checked again we do not need the code anymore. We have handled it by other way. We can directly merge this PR.

RajkiranBagal (approved):

assets/js/unminified/frontend.js
Nikschavan: Thank you for confirming @RajkiranBagal

vrundakansara (approved):

premanshup (approved):

PR #2634, opened 2021-04-16
AST-356 - refactor: revamp existing blog page CSS with grid based CSS

sagarb90:
### Description

- Revamp existing blog page CSS with grid-based CSS.
- https://wp-astra.atlassian.net/browse/AST-356
- Also, manage backward compatibility using the previous version filter.
- Removed all existing bootstrap CSS and mange the same layout using grid CSS.

### Screenshots
image

### Types of changes
-  Improvements

### How has this been tested?
-  Verify all the options from the customizer Blog > Blog/Archive section.

### Checklist:
- [x] My code is tested
- [x] I've included any necessary tests image


assets/css/unminified/main.css
```diff
@@ -1223,6 +1153,12 @@ body .ast-oembed-container * {
 }
 
 /*----------  Basic Layout  ----------*/
+:root {
```
Nikschavan, line 1156: Should we define the CSS variables in only one place - dynamic CSS, at the very top of it?

I imagine - if the user changes any values in customiser these variables will again be defined in the dynamic CSS so we will duplicate definition in customiser and static CSS?

Instead what I think is - have these set as default values for the customiser options and let it generate dynamic CSS based on default values by default and no definitions in static variables.

sagarb90: @Nikschavan Yes correct I have defined here temporary, We have already created a separate global variable dynamic CSS file for it in global variable change, After merging global variable changes these will be moving it to dynamic CSS.

PR #2635, opened 2021-04-16
AST-70 new: Added a filter to retrive 404 classes

rohitp-bsf:
### Description
- Added a new filter `astra_get_404_page_classes`

### Screenshots
- NA

### Types of changes
- Added a new filter only.

### How has this been tested?
- NA

### Checklist:
- [ ] My code is tested


template-parts/404/404-layout.php
```diff
@@ -9,8 +9,19 @@
...
+/**
+ * Filter to retrive classes.
+ *
+ * @return mixed|void
+ */
+function astra_get_404_layout_classes() {
+
+	$classes[] = 'ast-404-layout-1';
+	return apply_filters( 'astra_get_404_page_classes', $classes );
+}
```
premanshup, line 21: @rohitp-bsf We already have `astra_attr()` using which we can generate attributes and inside it dynamic filter is also there which we can use. I don't think there is need to add separate filter. We can just convert the below markup to `astra_attr()`. Please check.

premanshup (changes requested):

template-parts/404/404-layout.php
rohitp-bsf: Updated!

premanshup (approved):

PR #2656, opened 2021-04-23
revert: Reverted post__not_in parameter in WP_Query because it affects on number of posts count

imnavanath:
### Description
- As `post__not_in` parameter removed from WP_Query and managed within the loop with `if` condition
- This arises with a corner case - https://share.bsf.io/o0u6GpNR
- Post count is showing less than 1
- This happens because if we set total posts count 4 & order desc (just for getting the case, configs can be anything)
- If current active (opened) post is coming up within this 4 posts then obviously within the loop this will exclude & only 3 posts will shown up
- This is the drawback of removing `post__not_in` param from WP_Query & managed through `while loop > if` condition.
- That's why reverting this change.
- Performance measures for this -
- Without post__not_in - https://blackfire.io/profiles/8457a3e8-bb48-4c48-a7de-184c6a903b15/graph
- With post__not_in - https://blackfire.io/profiles/920f10d4-2e1e-472f-9914-18d2b4bb155c/graph

### Screenshots
- https://share.bsf.io/o0u6GpNR

### Types of changes
- Readding `post_not_in` parameter inside WP_Query.

### How has this been tested?
- Steps as per above scrrencast

### Checklist:
- [x] My code is tested


Nikschavan:
@imnavanath - would it work if you requests `n + 1` or `n + 2` number of posts and then select first `n` number of posts from the array?

imnavanath:
Closing this PR because already worked with alternate solution [here](https://github.com/brainstormforce/astra/pull/2657/)

PR #2657, opened 2021-04-26
refactor: WP_Query corrected if current post is one of the post from loop & which reduces post count by 1

imnavanath:
### Description
- As `post__not_in` parameter removed from WP_Query and managed within the loop with `if` condition
- This arises with a corner case - https://share.bsf.io/o0u6GpNR
- Post count is showing less than 1.
- This happens because if we set total posts count 4 & order desc (just for getting the case, configs can be anything)
- If current active (opened) post is coming up within this 4 posts then obviously within the loop this will exclude & only 3 posts will shown up
- This is the drawback of removing `post__not_in` param from WP_Query & managed through `while loop > if` condition.
- Improvement as per this comment - https://github.com/brainstormforce/astra/pull/2656#issuecomment-825800268

### Screenshots
- https://share.bsf.io/o0u6GpNR
- Font size preview not working - https://share.getcloudapp.com/9ZuxgAL8

### Types of changes
- Taken one extra post to cover the gap up of current active post (if that comes in loop previously)
- Managed post count & while loop as per post count

### How has this been tested?
- Steps as per above screencast

### Checklist:
- [x] My code is tested


inc/modules/related-posts/related-posts.php
```diff
@@ -22,14 +22,16 @@
 function astra_get_related_posts_by_query( $post_id ) {
 	$term_ids                  = array();
 	$current_post_type         = get_post_type( $post_id );
-	$related_posts_total_count = astra_get_option( 'related-posts-total-count', 2 );
+	$related_posts_total_count = (int) astra_get_option( 'related-posts-total-count', 2 );
+	// Taking one post extra in loop because if current post excluded from while loop then this extra one post will cover total post count. Apperently avoided 'post__not_in' from WP_Query.
+	$updated_total_posts_count = ( $related_posts_total_count >= 1 ) ? ( $related_posts_total_count + 1 ) : $related_posts_total_count;
```
Nikschavan, line 27: Why not simply do it `$related_posts_total_count + 1` instead of this tertiary condition?

Can you explain what this condition is doing?

inc/modules/related-posts/related-posts.php
```diff
@@ -22,14 +22,16 @@
 function astra_get_related_posts_by_query( $post_id ) {
 	$term_ids                  = array();
 	$current_post_type         = get_post_type( $post_id );
-	$related_posts_total_count = astra_get_option( 'related-posts-total-count', 2 );
+	$related_posts_total_count = (int) astra_get_option( 'related-posts-total-count', 2 );
```
Nikschavan, line 25:
```suggestion
	$related_posts_total_count = absint( astra_get_option( 'related-posts-total-count', 2 ) ) + 1;
```

Will this work?

inc/modules/related-posts/related-posts.php
```diff
@@ -259,8 +262,37 @@ function astra_get_related_posts() {
...
-		while ( $query_posts->have_posts() ) {
+		/**
+		 * To manage posts loop counter & total posts count has been managed here.
+		 *
+		 * # Loop count as per post set count.
+		 *
+		 * +-------+-------------------+---------------------------------+-----------------------------------------------------------+
+		 * | option  | expectation     | posts counter for further loop  | total posts on which our loop will execute                |
+		 * +-------+-------------------+---------------------------------+-----------------------------------------------------------+
+		 * | -1      | All posts       | -20                             | Assigned count of posts                                   |
```
Nikschavan, line 273: We should not allow `-1` or allow to fetch all the posts, It is even better to add a certain higher ceiling limit to number of posts that can be queried with this.

inc/modules/related-posts/related-posts.php
```diff
@@ -259,8 +262,37 @@ function astra_get_related_posts() {
 
 		do_action( 'astra_related_posts_loop_before' );
 
-		while ( $query_posts->have_posts() ) {
+		/**
+		 * To manage posts loop counter & total posts count has been managed here.
+		 *
+		 * # Loop count as per post set count.
+		 *
+		 * +-------+-------------------+---------------------------------+-----------------------------------------------------------+
```
Nikschavan, line 270: Instead of a table I think just a long paragraph with examples would be much easier to understand, IMO

inc/modules/related-posts/related-posts.php
imnavanath: We used tertiary condition because there were cases for this, we normally get 3 types values 
1. -1 = To load all posts
2. 0 = Zero count (alternate option is directly disable Related Posts section)
3. Any positive count = Number of posts
So `$related_posts_total_count + 1` this condition fails in -1 & 0 case, that's why tertiary used.

--

**What we applied logic here** - If posts count is -1 or 0 then `$updated_total_posts_count` will be same number (i.e. -1 or 0) on other side if it is any positive number then we will load +1 extra post to tackle post_not_in case.

--

After following https://github.com/brainstormforce/astra/pull/2657#discussion_r620154847 comment this will resolve I think, because we remove `-1` support & we will add slider control limit to 1-20.
So this extra code work will reduce.

Nikschavan: Right limit of 1-20 sounds good 👍

inc/modules/related-posts/related-posts.php
imnavanath: Surely will work 👍 I'll update & after resolve this comment.

imnavanath: Done thank you.

inc/modules/related-posts/related-posts.php
imnavanath: Agreed. Done. 👍

inc/modules/related-posts/related-posts.php
imnavanath: Setup 1-20 count limit 👍

inc/modules/related-posts/related-posts.php
imnavanath: True, 👍

premanshup (approved):

PR #2708, opened 2021-05-05
AST-445 - imprv: remove static CSS related to site logo and move it to dynamic conditionally

sagarb90:
### Description
 - Remove static CSS related to site logo and move it to dynamic conditionally.
 - https://wp-astra.atlassian.net/browse/AST-445
 
### Checklist:
- [x] My code is tested


inc/class-astra-dynamic-css.php
```diff
@@ -547,6 +547,16 @@ public static function return_output( $dynamic_css, $dynamic_css_filtered = '' )
 				);
 			}
 
+			if ( get_theme_mod( 'custom_logo' ) || is_customize_preview() ) {
```
premanshup, line 550: @sagarb90 Please verify the condition when normal logo is not added but logo is added from Transparent header or Sticky header or Page Header. 
1. Is this CSS needed for above types of logos?
2. If yes, then does this condition work properly with the these conditions?

premanshup (changes requested):

inc/class-astra-dynamic-css.php
premanshup: @sagarb90 Can you check the blackfire report and page load report how much difference does it make? Further, In theme, we can keep just transparent header condition. Sticky and Page Header conditions can go in addon, let me know if this will help us reduce the conditions?

sagarb90: @premanshup Yes I will check the report, but if we add conditions in add-on then we need again add same CSS then if     add-on is activated then this leads to duplicate CSS from theme and add-on

premanshup: @sagarb90 Yes, that will lead to duplicate CSS. Can we check the stats by adding all the condition? Are we getting any optimization or not? If not, we can leave this as it is.

sagarb90: Yes as there is only 2 CSS for logo which we have removed and adding these condition in PHP is no any impact on stats means this will not be showing any optimization.

premanshup (approved):

PR #2722, opened 2021-05-07
AST-452 - imprv: removed default loading of search component related CSS and load conditionally from dynamic CSS

sagarb90:
### Description
- Removed default loading of search component related CSS and load conditionally from dynamic CSS.
- We have removed only for New HFB and for Old HFB we kept as its as there is no way add conditional or need to to add in a separate place.
- https://wp-astra.atlassian.net/browse/AST-452

### Types of changes
- Improvements

### How has this been tested?
- Test search component related CSS in all devices.

### Checklist:
- [x] My code is tested


sass/site/_main.scss
```diff
@@ -7,7 +7,7 @@
 @import "variables-site/typography";
 
 /*----------  Site Normalize  ----------*/
-@import "site-normalize/site-normalize";
+@import "site-normalize/site-normalize-builder";
```
premanshup, line 10: @sagarb90 Can you please explain why we are loading new file called `site-normalize-builder`?

inc/builder/type/header/search/dynamic-css/dynamic.css.php
```diff
@@ -124,3 +153,176 @@ function astra_hb_search_dynamic_css( $dynamic_css, $dynamic_css_filtered = '' )
...
+ * @since x.x.x
+ */
+function astra_search_static_css() {
+	$search_css = '
+	.ast-header-break-point.ast-header-custom-item-inside .main-header-bar .main-header-bar-navigation .ast-search-icon {
+		display: none;
+	}
+	.ast-search-icon .astra-search-icon {
+		font-size: 1.3em;
+	}
```
premanshup, line 170: @sagarb90 This CSS is not removed from this code and still we are loading this in here. Any reason?

inc/builder/type/header/search/dynamic-css/dynamic.css.php
```diff
@@ -124,3 +153,176 @@ function astra_hb_search_dynamic_css( $dynamic_css, $dynamic_css_filtered = '' )
...
+			padding-right: 3em;
+			padding-left: unset;
+			right: -1em;
+			left: unset;
+		}
+		  
+		.site-header-section-left .ast-search-menu-icon.slide-search .search-form .search-field {
+			margin-left: unset;
+			margin-right: 10px;
+		}
```
premanshup, line 270: @sagarb90 Is this CSS duplicated with dynamic CSS - https://github.com/brainstormforce/astra/pull/2722/files#diff-1fa603c7b6bf95f4695a33ac8649f3b4445613784707b6b07e36b74bb1819e6aR68. Please check!

premanshup (changes requested):

imnavanath:
@sagarb90 facing this case where left-right spacing is missing https://share.bsf.io/X6ueywy6 (same thing I noticed in next-release as well) so probably not issue in this PR. But if we can fix this one in current PR would be much better.

sass/site/_main.scss
sagarb90: Reverted this as this is related to OLD HFB and which is already handle in this [PR  ](https://github.com/brainstormforce/astra/pull/2737)

inc/builder/type/header/search/dynamic-css/dynamic.css.php
sagarb90: This CSS should be removed from here I hade added here but this is common CSS in OLD and New HFB so I have by mistake added but not removed. and our search dynamic CSS is in builder folder so we can't add here

inc/builder/type/header/search/dynamic-css/dynamic.css.php
sagarb90: Yes this was duplicate in dynamic and removed

sagarb90:
> @sagarb90 facing this case where left-right spacing is missing https://share.bsf.io/X6ueywy6 (same thing I noticed in next-release as well) so probably not issue in this PR. But if we can fix this one in current PR would be much better.

I have check this in current and master branch as well this look correct at my end - https://share.getcloudapp.com/9ZudenWB

premanshup (approved):

PR #2778, opened 2021-05-19
Update customizer call as per web stories refactored code

pradeep910:
### Description
In the web stories plugin version 1.7.1, the code was refactored and plugin structure got changed. 
Updated web stories code as per the refactoring of web stories. See https://github.com/google/web-stories-wp/pull/7266/files

### Types of changes
Bug fix - web stories customizer disappeared due to recent web stories refactor changes. 

### Checklist:
- [x] My code is tested

### DIFF
--- a/inc/compatibility/class-astra-web-stories.php
+++ b/inc/compatibility/class-astra-web-stories.php
@@ -7,8 +7,6 @@
  * @package Astra
  */
 
-use Google\Web_Stories;
-
 /**
  * Astra Web_Stories Compatibility
  *
@@ -45,12 +43,12 @@ class Astra_Web_Stories {
 	 * @return void
 	 */
 	public function web_stories_embed() {
-		if ( ! function_exists( 'Google\Web_Stories\render_theme_stories' ) ) {
+		if ( ! function_exists( '\Google\Web_Stories\render_theme_stories' ) ) {
 			return;
 		}
 
 		// Embed web stories above header with pre-configured customizer settings.
-		Web_Stories\render_theme_stories();
+		\Google\Web_Stories\render_theme_stories();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


inc/compatibility/class-astra-web-stories.php
```diff
@@ -9,8 +9,12 @@
...
 use Google\Web_Stories;
 
-// If plugin - 'Google\Web_Stories' not exist then return.
-if ( ! class_exists( 'Google\Web_Stories\Customizer' ) ) {
+/**
+ * If plugin - 'Google\Web_Stories' not exist then return.
+ * Uses Web_Stories\Admin\Customizer per web stories 1.7.1 release.
+ * See https://github.com/google/web-stories-wp/pull/7266/files
+ */
+if ( ! class_exists( 'Google\Web_Stories\Admin\Customizer' ) ) {
```
swissspidy, line 17: I'd replace this with `function_exists( 'Google\Web_Stories\render_theme_stories' )` in case the `Customizer` class is refactored further on our side. I'd not use that class directly.

inc/compatibility/class-astra-web-stories.php
```diff
@@ -69,7 +73,8 @@ public function web_stories_embed() {
 	 * @return String Generated dynamic CSS for Heading Colors.
 	 */
 	public function web_stories_css( $dynamic_css, $dynamic_css_filtered = '' ) {
-		$options = get_option( Web_Stories\Customizer::STORY_OPTION );
+		// Use Web_Stories\Admin\Customizer per web stories 1.7.1 release. See https://github.com/google/web-stories-wp/pull/7266/files
+		$options = get_option( Web_Stories\Admin\Customizer::STORY_OPTION );
```
swissspidy, line 77: An alternative and refactoring-safe approach would be:

```php
		$options = get_option( 'web_stories_customizer_settings' );
```

inc/compatibility/class-astra-web-stories.php
swissspidy: This check could also be moved to the `web_stories_css` method instead

pradeep910: Yes, makes sense. 👍

pradeep910: I figured we don't need class check at this point. It is too early for function check, so used this before embedding and outputting CSS.

inc/compatibility/class-astra-web-stories.php
pradeep910: Thanks this is better.

inc/compatibility/class-astra-web-stories.php
```diff
@@ -9,11 +9,6 @@
 
 use Google\Web_Stories;
```
swissspidy, line 10: Now unused I think?

pradeep910: Yes and updated function call too for consistency.

Nikschavan (approved):

Nikschavan:
Thank you for checking into this guys :)
We will try to send this update out ASAP.

vrundakansara (approved):

PR #2780, opened 2021-05-20
AST-499 - feat: added SVG logo height option to resolve the CLS issue

sagarb90:
### Description
- Added SVG logo height property to resolve the CLS issue
- SVG logo responsive height will be calculated based on logo width at the background and then we save this height on same responsive logo settings as `{desktop: 127, tablet: "", mobile: "", desktop-svg-height: 0,tablet-svg-height: 0,mobile-svg-height: 0}`.
- Added setTimeout function to resolve the issue of getting wrong height calculation while changing the width using slider.
- For existing Users to get the height for the logo they need to edit the normal or transparent logo width.

### Types of changes
- New feature (non-breaking change which adds functionality) 

### Checklist:
- [x] My code is tested


inc/customizer/extend-custom-controls/src/control-customizer.js
```diff
@@ -72,6 +72,76 @@
 			});
 		});
 
+		/**
+		 * Trigger on ast-header-responsive-logo-width change.
+		 */
+		 api('astra-settings[ast-header-responsive-logo-width]', function (value) {
+			value.bind(function (currVal) {
+
+				let customizer_preview_container =  document.getElementById('customize-preview')
```
premanshup, line 81:
```suggestion
				const customizerPreviewContainer =  document.getElementById('customize-preview');
```

inc/customizer/extend-custom-controls/src/control-customizer.js
```diff
@@ -72,6 +72,76 @@
...
+				let iframe 						 = customizer_preview_container.getElementsByTagName('iframe')[0]
```
premanshup, line 82: @sagarb90 Can we change the selector from TagName to maybe ID? There can be multiple iframes and in this case, it can target some other iframe.

inc/customizer/extend-custom-controls/src/control-customizer.js
```diff
@@ -72,6 +72,76 @@
...
+				let htmlContent 				 = iframe.contentDocument || iframe.contentWindow.document;
+				
+				setTimeout(function () { 
+					if( htmlContent.querySelector('.astra-logo-svg') ) {
+						var existingValues 	    = api('astra-settings[ast-header-responsive-logo-width]').get();
```
premanshup, line 87:
```suggestion
						let existingValues 	    = api('astra-settings[ast-header-responsive-logo-width]').get();
```

inc/customizer/extend-custom-controls/src/control-customizer.js
```diff
@@ -72,6 +72,76 @@
...
+							case 'tablet':
+								existingValues['tablet-svg-height']  = mobileTabletHeight;
+								break;
+							case 'mobile':
+								existingValues['mobile-svg-height']  = mobileTabletHeight;
+								break;
+							default:
+								break;
+						}
+						api('astra-settings[ast-header-responsive-logo-width]').set( existingValues );
```
premanshup, line 104: @sagarb90 Instead of updating the existing array option for the same logo width, can we add hidden setting for SVG height separately?

premanshup (changes requested):

inc/customizer/extend-custom-controls/src/control-customizer.js
sagarb90: @premanshup Yes I was implement first solution by adding hidden inputs but it has two drawbacks
1 . We Need to add responsive hidden control and as their is currently ast-hidden is non-responsive, I had added responsive hidden but facing some problem with updating responsive values as it get saved in JS settings but it does not get saved in DB `astra-settings` I tried all the possible ways also discuss with rohit as well but cant resolved it.
2. We need to register 2 more controls one for normal logo and other for transparent logo width  this will increase the total controls size and again some PHP code as well.

So I think this is best and optimize solution for this.

inc/customizer/extend-custom-controls/src/control-customizer.js
sagarb90: @premanshup Yes but as I don't see any unique ID or class in the iframe tag so I used this 
ref- https://share.getcloudapp.com/5zuYprpZ

premanshup: @sagarb90 Can we target it using tagName? I see there is name attr in the iFrame.

inc/customizer/extend-custom-controls/src/control-customizer.js
premanshup: @sagarb90 Okay, right considering the array size, I agree, we should go with this. Please make sure to check any backward compatibility cases. Resolving this thread.

inc/customizer/extend-custom-controls/src/control-customizer.js
sagarb90: @premanshup No we can't use that name attr as it's value is dynamic and change sometime `customize-preview-0` and `customize-preview-2`.
https://share.getcloudapp.com/Kou4Gn2r

inc/customizer/extend-custom-controls/src/control-customizer.js
sagarb90: Done

inc/customizer/extend-custom-controls/src/control-customizer.js
premanshup: @sagarb90 Okay!

inc/customizer/extend-custom-controls/src/control-customizer.js
sagarb90: Done

premanshup (dismissed):

abhishek-bsf (approved):

premanshup (approved):

PR #2856, opened 2021-06-12
Feat: Theme Check - Automated GitHub action

premanshup:
### Description

Feature: Theme check using GitHub action and WP CLI Package https://github.com/anhskohbo/wp-cli-themecheck

### Screenshots

Image+2021-06-12+at+5 58 13+PM

### Types of changes

### How has this been tested?

Tested by running action for theme check the reports

### Theme Check Passed Report:

Image+2021-06-12+at+5 53 20+PM

### Theme Check Failed Report:

Image+2021-06-12+at+5 55 48+PM

### Checklist:
- [x] My code is tested


imnavanath (approved):

sagarb90 (approved):

imnavanath (approved):

Nikschavan (changes requested):
I think instead of introducing a new environment we should integrate this with the existing `local-env`

One thing we will have to change in `local-env` is instead of mounting the whole workspace, we should "build" the package and then only mount the built directory.

We will also need a way to install extra plugins, maybe we can do this by passing name of the shell script which will be run from the [install-wordpress.sh](https://github.com/brainstormforce/astra/blob/master/bin/local-env/install-wordpress.sh)

Let me know and we can get on a call to discuss this more.

premanshup:
> I think instead of introducing a new environment we should integrate this with the existing `local-env`
> 
> One thing we will have to change in `local-env` is instead of mounting the whole workspace, we should "build" the package and then only mount the built directory.
> 
> We will also need a way to install extra plugins, maybe we can do this by passing name of the shell script which will be run from the [install-wordpress.sh](https://github.com/brainstormforce/astra/blob/master/bin/local-env/install-wordpress.sh)
> 
> Let me know and we can get on a call to discuss this more.

@Nikschavan I have scheduled a call for today. Let's discuss this.

bin/local-env/theme-check.sh
```diff
@@ -0,0 +1,16 @@
...
+set -e
+
+# Include useful functions
+. "$(dirname "$0")/includes.sh"
+
+# Install theme-check package.
+wp package install anhskohbo/wp-cli-themecheck
+
+# Install theme-check plugin.
+wp plugin install theme-check --version=20200922.1 --activate
```
imnavanath, line 13: Just curious @premanshup is there any way, we target their stable released version from WP here instead of hardcoded `20200922.1` or might be targeting their [gitthub repo](https://github.com/WordPress/theme-check) > master version?

Nikschavan: This was added I think because there was a fatal error introduced with the latest version of theme check plugin and  `anhskohbo/wp-cli-themecheck` action. @premanshup do check if this is fixed and if not report the error.

It's better to revert this to master so that this gets always updated.

premanshup: @Nikschavan Yes, It is reported here - https://github.com/anhskohbo/wp-cli-themecheck/issues/6. Will keep watching this.

Nikschavan:
Closing in favour of #2911

PR #2863, opened 2021-06-15
AST-579 Bug in searchform.php translation

murshed:
Bug in searchform.php translation


Nikschavan (dismissed):

premanshup (dismissed):

imnavanath (dismissed):

sagarb90 (dismissed):

premanshup:
@murshed Thank you for the PR, this will be released in the version v3.7.0.

murshed:
I have modified the code.
If we added 'astra" on the line the translation does not work! Video- https://streamable.com/sj2utg

searchform.php
```diff
@@ -21,6 +21,6 @@
 			</button>
 		<?php } ?>
 	</label>
-	<input type="submit" class="search-submit" value="Search">
+	<input type="submit" class="search-submit" value="<?php echo esc_attr_x( 'Search', 'submit button' ); ?>" />
```
premanshup, line 24:
```suggestion
	<input type="submit" class="search-submit" value="<?php echo esc_attr_x( 'Search', 'submit button', 'astra' ); ?>" />
```

premanshup (changes requested):

premanshup:
> I have modified the code.
> If we added 'astra" on the line the translation does not work! Video- https://streamable.com/sj2utg

@murshed For some reason the link you shared is not working and I think it can still be translated using Loco Translate plugin. Since it will not appear the translatable strings.

PR #2865, opened 2021-06-15
AST-568 - imprv: update the default values of  blog/archive page settings for new default blog post design

sagarb90:
### Description
- https://wp-astra.atlassian.net/browse/AST-568
- Update the default values of  blog/archive page settings for new default blog post design as per designers inputs

### Types of changes
- Improvements

### Checklist:
- [x] My code is tested


inc/core/class-astra-theme-options.php
```diff
@@ -450,7 +450,7 @@ public static function defaults() {
 						'mobile-unit'  => 'px',
 					),
 					'font-size-h6'                         => array(
-						'desktop'      => 15,
+						'desktop'      => astra_apply_new_default_blog_values() ? 16 : 15,
```
imnavanath, line 453: Instead of calling this function again and again can we call it once in construct and save the value in a static variable and we use that in the above cases?

sagarb90: Right! I have updated the same.

PR #3049, opened 2021-07-28
AST-694 - Removing widget design options in WP-5.8 setup

imnavanath:
### Description
- We are going to remove color, typography options from the widget block editor from the customizer.
- Task - https://wp-astra.atlassian.net/browse/AST-694
- Dev site - http://ast-dev.xka9ibwowp-jqp3vd915350.p.runcloud.link/local-google-fonts/wp-admin/customize.php (checked as old user here and on local tested with deleting astra-settings as new user)
- Filter to use - `add_filter( 'astra_remove_widget_design_options', '__return_false' );`

### Types of changes
- Kind of improvement

### How has this been tested?
- Details added in the above card

### Checklist:
- [x] My code is tested


changelog.txt
```diff
@@ -1,3 +1,6 @@
+v3.7.1 (Unreleased)
+- Improvement: Builder - Removed design specific options from widgets in WordPress 5.8.
```
premanshup, line 2:
```suggestion
- Deprecated: Header footer widgets - Deprecated design specific sticky header options for better compatibility with WordPress 5.8.
```

premanshup (changes requested):

changelog.txt
imnavanath: Done 👍

inc/theme-update/class-astra-theme-background-updater.php
```diff
@@ -108,6 +108,9 @@ class Astra_Theme_Background_Updater {
 				'astra_remove_elementor_toc_margin',
 				'astra_clear_all_assets_cache',
 			),
+			'3.7.1' => array(
```
premanshup, line 111:
```suggestion
			'3.6.8' => array(
```

inc/extras.php
```diff
@@ -685,3 +685,40 @@ function astra_can_remove_elementor_toc_margin_space() {
...
+
+	// True -> Hide widget sections, False -> Display widget sections.
+	$is_widget_design_sections_hidden = true;
+
+	if ( ! $astra_settings['remove-widget-design-options'] ) {
+		// For old users we will show widget design options by anyways.
+		$is_widget_design_sections_hidden = false;
+	} else {
+		// Considering the user is new now.
+		if ( astra_get_option( 'remove-widget-design-options' ) ) {
```
premanshup, line 711: @imnavanath Check if we can optimize this by reducing the DB call and using the above DB call.

functions.php
```diff
@@ -15,7 +15,7 @@
 /**
  * Define Constants
  */
-define( 'ASTRA_THEME_VERSION', '3.6.7' );
+define( 'ASTRA_THEME_VERSION', '3.7.1' );
```
premanshup:
```suggestion
define( 'ASTRA_THEME_VERSION', '3.6.8' );
```

premanshup (changes requested):

inc/theme-update/class-astra-theme-background-updater.php
imnavanath: Done

functions.php
imnavanath: Done

inc/extras.php
imnavanath: Please review once now, used isset() & new var for avoided astra_get_option() in condition.

premanshup (approved):

inc/extras.php
```diff
@@ -685,3 +685,40 @@ function astra_can_remove_elementor_toc_margin_space() {
...
+ *
+ * @since x.x.x
+ * @return boolean
+ */
+function astra_remove_widget_design_options() {
+	$astra_settings               = get_option( ASTRA_THEME_SETTINGS );
+	$remove_widget_design_options = isset( $astra_settings['remove-widget-design-options'] ) ? false : true;
+
+	// True -> Hide widget sections, False -> Display widget sections.
+	$is_widget_design_sections_hidden = true;
```
sagarb90, line 704: Instead of introducing a new variable, we can also use `$remove_widget_design_options` and update the flag(true/false) Also below if condition is also not required in that case.

imnavanath: we can eliminate first if condition and only 3 `if-elseif-else` conditions will remain. But then for every old user second elseif condition satisfy and it will hide those options. Whereas it needed to be visible anyhow for old users.

sagarb90: Okay Can do like this then,
return early and avoid nesting of code.
```suggestion
	if ( ! $remove_widget_design_options ) {
		// For old users we will show widget design options by anyways.
		return apply_filters( 'astra_remove_widget_design_options', $is_widget_design_sections_hidden );
	}
	// Considering the user is new now.
	if ( isset( $astra_settings['remove-widget-design-options'] ) && $astra_settings['remove-widget-design-options'] ) {
		// User was on WP-5.8 lesser version previously and he may update their WordPress to 5.8 in future. So we display the options in this case.
		$is_widget_design_sections_hidden = false;
	} elseif ( astra_has_widgets_block_editor() ) {
		// User is new & having block widgets active. So we will hide those options.
		$is_widget_design_sections_hidden = true;
	} else {
			// Setting up flag because user is on lesser WP versions and may update WP to 5.8.
			astra_update_option( 'remove-widget-design-options', true );
	}
	

	return apply_filters( 'astra_remove_widget_design_options', $is_widget_design_sections_hidden );
```

imnavanath: Done 👍

PR #3085, opened 2021-08-09
AST-582- imprv: Clear asset cache of the theme and addon on every new update of the theme

sagarb90:
### Description
- Clear asset cache of the theme and addon on every new update of the theme
- https://wp-astra.atlassian.net/browse/AST-582
- Test with enable file generation and update the theme, It should clear theme and add-on asset cache( from Upload directory ) 

### Types of changes
- Improvements

### Checklist:
- [x] My code is tested


inc/theme-update/class-astra-theme-background-updater.php
```diff
@@ -397,6 +392,29 @@ public static function update_db_version( $version = null ) {
 			delete_transient( 'astra-addon-db-migrated' );
 
 			do_action( 'astra_theme_update_after' );
+			
+			// This will clear Astra and Astra Pro asset cache on every new update.
+			self::astra_clear_all_assets_cache();
```
premanshup, line 397:
```suggestion
			astra_clear_all_assets_cache();
```
@sagarb90 Instead of creating a new function, can we call this existing function directly?

premanshup (changes requested):

inc/theme-update/class-astra-theme-background-updater.php
sagarb90: @premanshup Yes we can use the existing function I thought we have functions in `astra-update-functions.php` file which are called in db update process.

changelog.txt
```diff
@@ -1,5 +1,6 @@
 v3.6.8 (Unreleased)
 - Improvement: Added WPML translation support for all customizer's strings through wpml-config.xml file.
+- Improvement: Clear asset cache of theme and addon on every new update of theme.
```
premanshup, line 3: @sagarb90 This can go under `v3.6.9`

premanshup (approved):

inc/theme-update/class-astra-theme-background-updater.php
```diff
@@ -366,7 +366,6 @@ public static function update_db_version( $version = null ) {
 
 			// If equals then return.
 			if ( version_compare( $saved_version, ASTRA_THEME_VERSION, '=' ) ) {
-				do_action( 'astra_theme_update_after' );
```
imnavanath, line 369: We removed this because already in this function same action exists and also this is not right position of this action, right? or any other reason is there?

sagarb90: @imnavanath Yes This is not the right position to add action because this is will run on every page refresh. No other reason

imnavanath (approved):

inc/theme-update/astra-update-functions.php
```diff
@@ -3074,15 +3074,17 @@ function astra_transparent_header_default_value() {
  * @return void.
  */
 function astra_clear_all_assets_cache() {
```
imnavanath, line 3076: a thought - can we move this function from this batch functions specific file, we can move this in common-functions / helper functions itself. Same for addon as well.

sagarb90: Yes I was also thinking the same but we have reused this function. otherwise, we have to remove it from here and update the existing function calls as well, Also this is also same job as batch function i.e run after each update. So we can keep here

changelog.txt
sagarb90: Yes this is done

inc/theme-update/astra-update-functions.php
imnavanath: IMO, as soon as we deploy this code, cache clears from every site so no longer needed those old batch function calls (from previous versions) we can remove those and keep this one only. 

Your thought also @premanshup ??

premanshup: @imnavanath Yes, It will work with each update, @sagarb90 can we remove the function calls since this new logic will work after each update regardless of the version number?

sagarb90: @premanshup Yes I will remove those.

sagarb90: I have removed please review it once.

premanshup (approved):

PR #3130, opened 2021-08-17
AST-494 - imprv: Move sticky option to sticky header

sagarb90:
### Description
- Move sticky options of individual components in the sticky header section.
- Showrespective component's sticky option only if they are placed in any of the header sections with sticky enabled.
- We have added condition show and hide sticky headers component's controls logic in the theme.

### Types of changes
- Improvements

### Checklist:
- [x] My code is tested


premanshup:
@sagarb90 Please add description to the PR.

inc/customizer/extend-custom-controls/src/index.js
```diff
@@ -34,6 +34,7 @@ import { colorGroupControl } from './color-group/control';
 import { selectorControl } from './selector/control';
 import { boxShadowControl }  from './box-shadow/control.js';
 import { responsiveToggleControl } from './responsive-toggle-control/control';
+import { controlTitleControl } from './controls-title/control';
```
premanshup, line 37: @sagarb90 We already have 
`import { headingControl } from './heading/control';`
Can you please check if this can be used and if we can avoid adding new control?

premanshup (changes requested):

inc/customizer/extend-custom-controls/src/index.js
sagarb90: @premanshup No, we can't use this as this is different control - https://share.getcloudapp.com/jku46JpA
And we need this new control because we have a divider context where we can add titles but some controls are shown on a conditional basis so we can't add title in tit so we introduce this new control for heading.

premanshup: @sagarb90 I see! Can we use divider? Just like we have used it here - https://share.bsf.io/mXuKzZq4. Please check.

inc/core/builder/class-astra-builder-options.php
```diff
@@ -1119,8 +1119,6 @@ function prepare_button_defaults( $defaults, $index ) {
 
 	$defaults[ 'section-hb-button-' . $index . '-padding' ]   = Astra_Builder_Helper::$default_responsive_spacing;
 	$defaults[ 'section-hb-button-' . $index . '-margin' ]    = Astra_Builder_Helper::$default_responsive_spacing;
-	$defaults[ 'sticky-header-button' . $index . '-padding' ] = Astra_Builder_Helper::$default_responsive_spacing;
```
sagarb90, line 1122: This is removed and move it to addon as this is related to sticky and need to load from the addon

PR #2694, opened 2021-05-04
AST-431 fix: menu close issue when resize event trigger on keyboard display

shubhw12:
### Description
I have added a condition check before we trigger the menu toggle logic in the window.resize event callback.

### Types of changes
Bugfix (non-breaking change which fixes an issue)

### How has this been tested?
Checked if the closes after resize?
Checked if the menu closes if the active element is input field.

### Checklist:
- [x] My code is tested
- [x] My code passes the PHPCS tests
- [x] My code follows accessibility standards 
- [x] My code has proper inline documentation 
- [x] I've included any necessary tests 
- [x] I've included developer documentation 
- [x] I've added proper labels to this pull request

PR #3140, opened 2021-08-19
AST-598 fix: The submenu is not expanding when Flyout menu is used in…

shubhw12:
### Description
The submenu is not expanding when the Flyout menu is used in browser resizing.

### Screenshots

https://share.getcloudapp.com/yAuDO7yB

### Types of changes

 Bugfix (non-breaking change which fixes an issue)

### How has this been tested?

Checked the working of the changes for all the header types of off-canvas menu.

### Checklist:
- [ ] My code is tested


assets/js/unminified/frontend.js
```diff
@@ -450,9 +450,8 @@ var astraTriggerEvent = function astraTriggerEvent( el, typeArg ) {
 
 		updateHeaderBreakPoint();
 
-		if ( 'dropdown' === mobileHeaderType ) {
-			AstraToggleSetup();
-		}
+		AstraToggleSetup();
```
premanshup, line 453: @shubhw12 This condition was added for this fix https://github.com/brainstormforce/astra/pull/2694/files
I think it would be better to add OR condition here for the 
'flyout' === mobileHeaderType 
Flyout case as well otherwise the last issue may appear again.
Please check.

premanshup (changes requested):

assets/js/unminified/frontend.js
shubhw12: This issue is there for the full-screen menu as well. then there is no point in adding conditions.

I have checked the keyboard case the issue does not appear again.

premanshup: Okay if previous issue is also fixed then we can go ahead.

premanshup (approved):

PR #3194, opened 2021-09-01
Feat: Replace Global Color Palette 'editor-color-palette' to theme.json

premanshup:
### Description

This PR introduced theme.json in the theme. Global Color Palette for GB editor works on the basis of theme.json

### Screenshots

Image 2021-09-01 at 6 03 12 PM

### Types of changes

### How has this been tested?

Tested by changing the global color from customizer and applying the colors to the text/background.

### Checklist:
- [ ] My code is tested


premanshup:
Checking the failed GitHub checks

premanshup:
> Checking the failed GitHub checks

Static analysis is passed, Size check will get passed once we will pull the `release-candidate` in this as it is fix there. @patilvikasj Can you please pull the latest `release-candidate` in `release/v3.7.0-beta.2`

patilvikasj (approved):

inc/customizer/configurations/layout/class-astra-site-container-layout-configs.php
```diff
@@ -132,6 +132,9 @@ public function register_configuration( $configurations, $wp_customize ) {
 				),
 			);
 
+			/** @psalm-suppress UndefinedClass */ // phpcs:ignore Generic.Commenting.DocComment.MissingShort
+			$section_content_bg_obj = ( defined( 'ASTRA_EXT_VER' ) && Astra_Ext_Extension::is_active( 'colors-and-background' ) ) ? 'section-colors-body' : 'section-colors-background';
```
Nikschavan, line 136: Does adding `Class_Exists` fix the psalm complaint?
Also we should add class_exists instead of checking the constant anyway AST-737

inc/core/class-gutenberg-editor-css.php
```diff
@@ -964,6 +967,7 @@ public static function get_css() {
 
 				/** @psalm-suppress InvalidArgument */ // phpcs:ignore Generic.Commenting.DocComment.MissingShort
 				$css .= astra_parse_css( $boxed_container_tablet, '', astra_get_tablet_breakpoint() );
+				/** @psalm-suppress InvalidArgument */ // phpcs:ignore Generic.Commenting.DocComment.MissingShort
```
Nikschavan, line 970: Does `@psalm-suppress` take inline comments format `//`>

inc/class-astra-global-palette.php
```diff
@@ -24,12 +24,75 @@ class Astra_Global_Palette {
...
+	 *
+	 * @since x.x.x
+	 * @return void
+	 */
+	public function support_editor_color_palette() {
+		$global_palette = astra_get_option( 'global-color-palette' );
+		$editor_palette = $this->format_global_palette( $global_palette );
+
+		// Editor Color Palette.
+		add_theme_support( 'editor-color-palette', $editor_palette );
```
Nikschavan, line 59: Comment seems superfluous, maybe drop the comment or if there is any better description that can be added here.

inc/class-astra-global-palette.php
```diff
@@ -24,12 +24,75 @@ class Astra_Global_Palette {
...
+	}
+
+	/**
+	 * Format color palette data required to pass for Gutenberg palette.
+	 *
```
Nikschavan, line 64: Not a requirement: Do try to describe the functions a bit more, in this case - what is the actual difference in how we store it and how Gutenberg uses it.

Also, just wondering - this data formatting will be done on each page loads, is there any way to avoid this?
maybe storing the data as gutenberg requires? (not saying we should do it FYI but worth discussing)

inc/class-astra-global-palette.php
```diff
@@ -24,12 +24,75 @@ class Astra_Global_Palette {
...
 	public function __construct() {
-		add_action( 'after_setup_theme', array( $this, 'support_editor_color_palette' ) );
+		/**
+		 * Adding backward for theme.json file.
+		 *
+		 * If theme.json is not present in the child theme load Global Color Palette in the editor using add_theme_support( 'editor-color-palette', $editor_palette );.
+		 *
+		 * This is a known issue in Gutenberg - If theme.json is not present in the child theme, it does fallback to the parent theme's theme.json file.
+		 *
+		 * This will be fixed in the future updates of WordPress/Gutenberg.
```
Nikschavan, line 34:
```suggestion
		 * If theme.json is not present in the child theme load Global Color Palette in the editor using add_theme_support( 'editor-color-palette', $editor_palette );.
		 * This is a known issue in Gutenberg - If theme.json is not present in the child theme, it does fallback to the parent theme's theme.json file.
		 * This will be fixed in the future updates of WordPress/Gutenberg.
```

Removing extra lines in the comment description.

inc/class-astra-global-palette.php
```diff
@@ -24,12 +24,75 @@ class Astra_Global_Palette {
 	 * @since x.x.x
 	 */
 	public function __construct() {
-		add_action( 'after_setup_theme', array( $this, 'support_editor_color_palette' ) );
+		/**
+		 * Adding backward for theme.json file.
```
Nikschavan, line 28: Maybe word it - `Support for overriding theme.json from the child theme` instead?

Nikschavan (changes requested):

inc/customizer/configurations/layout/class-astra-site-container-layout-configs.php
premanshup: @Nikschavan It works with `class_exists` check. I tried it with `method_exists` function but psalm was throwing error for this. I will update it to class_exists.

inc/core/class-gutenberg-editor-css.php
premanshup: @Nikschavan If we add it as multi doc comment then it shows error in PHPCS about missing short. If we add short description then psalm shows error. This is the only way I could find to make them both work.

inc/customizer/configurations/layout/class-astra-site-container-layout-configs.php
premanshup: Updated!

inc/class-astra-global-palette.php
premanshup: Right! Removed!

inc/class-astra-global-palette.php
premanshup: Done!

inc/class-astra-global-palette.php
premanshup: Updated!

inc/class-astra-global-palette.php
premanshup: @patilvikasj Please check this, I think we can read this from theme.json as well.

patilvikasj: @premanshup We are reading this from theme.json itself. This is just a fallback function which will be called in the case where the user is using a child theme and WP does not use the parent theme theme.json as a fallback.

patilvikasj: Once WP uses parent theme.json as a fallback. We will deprecate/remove this code.

premanshup: @patilvikasj I have added a task for this as per the discussion, please check - https://wp-astra.atlassian.net/browse/GCP-155

PR #3826, opened 2021-11-25
AST-541 - CreativeWork Schema "image" size Error

GauravKhupse:
### Description
*Problem :* CreativeWork Schema Error on Structured Data Testing Tool when the Featured Images Size is set from the Customizer.

### Screenshots
 https://a.cl.ly/12ur2RJJ

### How has this been tested?
Added in the card.

### Checklist:
- [x] My code is tested


inc/core/class-astra-attr.php
```diff
@@ -72,7 +72,12 @@ public function astra_attr( $context, $attributes = array(), $args = array() ) {
 				if ( true === $value ) {
 					$output .= esc_html( $key ) . ' ';
 				} else {
-					$output .= sprintf( '%s="%s" ', esc_html( $key ), esc_attr( $value ) );
+					// Added image itemprop condition for schema markup.
+					if ( 'itemprop' === $key && 'image' === $value ) {
```
premanshup, line 76:
```suggestion
					if ( 'itemprop' === $key && 'image' === $value ) {
```
@GauravKhupse This condition should be dynamic. If in the future, we add any other string instead of image, this condition will fail. Please maybe try escaping the quotes if double qoutes are present in the string.

GauravKhupse: @premanshup We will discuss.

PR #3871, opened 2021-12-03
AST-1436 - Filter to disable all Meta Settings of Page/Post by default is not working in Gutenberg based metabox.

imnavanath:
### Description
- Filter to disable all meta settings is not working in react based metabox
- https://wpastra.com/docs/disable-all-meta-settings-of-page-post-by-default/

### Types of changes
- Bug fix

### How has this been tested?
- Add above doc's filter in child theme
- Verify on post/page, the all default disable checkboxes should enable with filter

### Checklist:
- [x] My code is tested


GauravKhupse (approved):

sagarb90:
@imnavanath We do not just set the values of fields, in this case, values will not be saved in DB, To save updated values in DB you need to also call this `props.setMetaFieldValue( val, value['key'] );`  on page reload to save the default values in DB;

sagarb90 (approved):

PR #4108, opened 2022-02-02
AST-561 - Margin issue for single shop page with the Full width & Contained Layout

nathanrodrigues2111:
### Description
Adds margin on top of woocommerce single product if it's a new user. For old users, if the user wants to change then he should use a filter provided to change the condition.

### Screenshots
https://d.pr/i/Qqrgt6

### Types of changes

### How has this been tested?
- Checked by setting up settings on an old Astra build to simulate old users and then updated to the latest version. ( Result: Margin should not be applied )
- For new users margin should be applied so checked it by setting my local as a new user.

### Checklist:
- [x] My code is tested


inc/class-astra-dynamic-css.php
```diff
@@ -1061,8 +1061,12 @@ public static function return_output( $dynamic_css, $dynamic_css_filtered = '' )
 					'1200'
 				);
 
+				// Checks if value is set and adds the condition accordingly.
+				$astra_condition_margin_flag = astra_get_option( 'apply-woo-product-single-margin', false );
+				$astra_condition_margin      = $astra_condition_margin_flag ? ( is_singular() ) : ( is_singular() && ! is_product() );
```
imnavanath, line 1066: @nathanrodrigues2111 `is_product()` will throw fatal error if WC is not installed, you can add check before that like class_exist WooComerce or funcion_exist is_product.

PR #4389, opened 2022-05-31
AST-1895 - PHP Notice - Array to string conversion on widget.php page.

nileshc-bsf:
### Description
Error message on the Widgets page

### Screenshots
https://share.bsf.io/p9uOxeKQ

### Types of changes

Bug fix (non-breaking change which fixes an issue) 

### How has this been tested?
- Enable the debug log
- Navigate to the Appearance => Widgets (notice will be displayed)
(On our end not able to replicate it checked on PHP7.4 but on users end it is showing)

### Checklist:
- [ ] My code is tested


inc/core/class-astra-enqueue-scripts.php
```diff
@@ -94,7 +94,7 @@ public function admin_body_class( $classes ) {
 			}
 
 			if ( ( ( 'post-new.php' == $pagenow || 'post.php' == $pagenow ) && ( defined( 'ASTRA_ADVANCED_HOOKS_POST_TYPE' ) && ASTRA_ADVANCED_HOOKS_POST_TYPE == $screen->post_type ) ) || 'widgets.php' == $pagenow ) {
-				return;
+				return '';
```
imnavanath, line 97:
```suggestion
				return $classes;
```

imnavanath: If we return blank then it will override previous added classes with empty string, that can be conflicting case. So we can directly return $classes from here, and won't add many more upcoming.

nileshc-bsf: Correct I was thinking the same, fixed it. 👍

imnavanath (approved):

PR #4527, opened 2022-08-09
v3.9.0 Fabcebook & from support ticket reported queries

yssambare12:
### Checklist:
- [ ] My code is tested


header.php
```diff
@@ -20,7 +20,7 @@
 <head>
 <?php astra_head_top(); ?>
 <meta charset="<?php bloginfo( 'charset' ); ?>">
-<meta name="viewport" content="width=device-width, initial-scale=1">
+<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
```
imnavanath, line 23: https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag

Any value less that 3 will cause AXE issue as default is 10.

yssambare12: > https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag
> 
> Any value less that 3 will cause AXE issue as default is 10.

Ok I will check this.

PR #4633, opened 2022-09-23
AST-2231 - Gutenberg editor padding not applied for heading blocks issue 

neerajmasai:
### Description

- BugFix: There should not be any padding, and it should look the same as in backend editor.
### Screenshots

Test Screencast: [Screencast at September 23rd 2022 - 2.mp4](https://d.pr/v/eb1X3X)

Note: please test with 3-4 different templates.
### Types of changes

- Gutenberg css specificity increased so that padding is applied correctly.

### How has this been tested?

- On a fresh install, installed Astra, Astra Addon, Spectra.
- Created new page with Gutenberg editor.
- Now take any heading block h1, h2, h3, and so on.
- Apply background color from the settings.
- Background color will be added to heading with appropriate padding.
- Check in frontend, same will be visible.

### Checklist:
- [x] My code is tested


sass/admin/block-editor-styles.scss
```diff
@@ -521,3 +521,13 @@ h6{
...
 	padding-left: 5px;
 }
+
+// Overriding padding which is inherited from wp-includes > block > heading > style.css
+h1.has-background.has-background,
+h2.has-background.has-background,
+h3.has-background.has-background,
+h4.has-background.has-background,
+h5.has-background.has-background,
+h6.has-background.has-background {
```
imnavanath, line 531: @neerajmasai is it redundant `.has-background.has-background` ?

neerajmasai: @imnavanath Actually, this is to increase specificity, we can also take two alternative solutions if needed, but imo this is best as per my research:

- Use a parent selector (but code will increase and it will be very specific, on some areas might not target)
- Use !important.

imnavanath: No I mean `.has-background.has-background` selector seems to be wrong, it is showing correct result (https://share.bsf.io/12uz4o6O) for now but same selector is applying twice, which IMO WordPress can fix in their next versions and after that we will again face this issue. So instead of one has-background you can use `.wp-block` like this - https://share.bsf.io/NQujoJz4

imnavanath, line 531:
```suggestion
h1.wp-block.has-background,
h2.wp-block.has-background,
h3.wp-block.has-background,
h4.wp-block.has-background,
h5.wp-block.has-background,
h6.wp-block.has-background {
```

neerajmasai: @imnavanath That indeed seems like the best approach, thanks, let me test that out.

imnavanath (approved):

PR #4802, opened 2022-12-02
AST-2284 - Sortable payment section support

nathanrodrigues2111:
### Description
Jira : https://brainstormforce.atlassian.net/browse/AST-2284

### Screenshots
https://d.pr/i/WWgo9X

### Types of changes

### How has this been tested?
- Checked when some options are selected for payment 
- Checked backward 
- Checked when both custom and premade are added at onces
- Checked with grey and inherit color 

### Checklist:
- [x] My code is tested


neerajmasai (approved):

neerajmasai:
@nathanrodrigues2111 Please update the changelog.

inc/theme-update/astra-update-functions.php
```diff
@@ -1346,3 +1346,110 @@ function astra_theme_background_updater_4_0_0() {
...
+ * @return void
+ */
+function astra_theme_background_updater_4_0_1() {
+
+	$theme_options = get_option( 'astra-settings', array() );
+
+	$current_payment_list = array();
+	$old_payment_list     = isset( $theme_options['single-product-payment-list']['items'] ) ? $theme_options['single-product-payment-list']['items'] : array();
+
+	$visa_payment       = isset( $theme_options['single-product-payment-visa'] ) ? $theme_options['single-product-payment-visa'] : '';
```
imnavanath, line 1363: I think else part should be `true` instead empty string 🤔 
See the case once - 
1. Master version of Astra
2. Reset customizer
3. Just enable payment section from sortavble list (don't change anything for toggles)
4. So all payment options will start display on frontend as default is true = https://share.bsf.io/X6u5D8Z5
5. Now migrate to this branch

With the logic - "not every default option save into astra-settings" your condition won't pass, and it leads into missing icons after migration. 

Can you verify this once.

inc/theme-update/astra-update-functions.php
```diff
@@ -1346,3 +1346,110 @@ function astra_theme_background_updater_4_0_0() {
...
+
+	false !== $visa_payment ? array_push(
+		$current_payment_list,
+		array(
+			'id'      => 'item-visa',
+			'enabled' => true,
+			'source'  => 'icon',
+			'icon'    => 'cc-visa',
+			'image'   => '',
+			'label'   => 'Visa',
```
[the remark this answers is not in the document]
nathanrodrigues2111: fixed

inc/theme-update/astra-update-functions.php
nathanrodrigues2111: so for the first one which are enabled by default i am checking if its != false as the initial value of those are null / empty / "" , So that is the reason why i had to check for != false . The migration works fine @imnavanath

PR #5013, opened 2023-02-20
AST-2872 - Block Editor Padding Does not Apply to Custom Layout blocks

neerajmasai:
### Description

- The block editor paddings in the customizer are not applicable to the custom layout blocks. The paddings can be seen in the editor but no padding in the front.

- How to reproduce:
   - Add a custom layout with a group block and add the display rules
   - Now try to change the block editor padding in the customizer.

- JIRA - https://brainstormforce.atlassian.net/browse/AST-2872

There is no change on the padding in the front but the padding is applied in the editor.
### Screenshots

https://d.pr/v/lwOn17

### Types of changes

- Fix: Block Editor Padding Does not Apply to Custom Layout blocks
- Changed selector for css to work for custom layout group, columns and cover blocks.

### How has this been tested?

- Add a custom layout with a group block and add the display rules
- Now try to change the block editor padding in the customizer.
- Check frontend and customizer preview.

### Checklist:
- [x] My code is tested


inc/dynamic-css/block-editor-compatibility.php
```diff
@@ -256,7 +256,7 @@ function astra_load_modern_block_editor_ui( $dynamic_css ) {
 	$astra_full_stretched_selector  = $astra_block_editor_v2_ui ? '.ast-plain-container.ast-no-sidebar .entry-content > .alignfull' . $exclude_uagb_selector . ', .ast-page-builder-template .ast-no-sidebar .entry-content > .alignfull' . $exclude_uagb_selector : '.ast-plain-container.ast-no-sidebar .entry-content .alignfull' . $exclude_uagb_selector . ', .ast-page-builder-template .ast-no-sidebar .entry-content .alignfull' . $exclude_uagb_selector;
 
 	$dynamic_css .= '
-	.entry-content > .wp-block-group, .entry-content > .wp-block-cover, .entry-content > .wp-block-columns {
+	.entry-content .wp-block-group, .entry-content .wp-block-cover, .entry-content .wp-block-columns {
```
nathanrodrigues2111, line 259: Did you check why this kind of specificity was added before ( > ) ? As this could be added to fix something else @neerajmasai

nathanrodrigues2111, line 259: Did you check why this kind of specificity was added before ( > ) ? As this could be added to fix something else @neerajmasai

neerajmasai: @nathanrodrigues2111  Yes, I think it was added because the group, cover and column blocks were direct children of `.entry-content` https://d.pr/i/kdEkak

However, in case of Custom layouts we are adding another child class to the existing hierarchy because of which the child combinator css is not working for custom layouts  https://d.pr/i/HsgY4h 

Therefore, I used the ancestor combinator instead to apply the css commonly for group, cover, column blocks for general case as well as custom layouts.

nathanrodrigues2111: okay @neerajmasai

nathanrodrigues2111 (approved):

yssambare12 (approved):

inc/dynamic-css/block-editor-compatibility.php
imnavanath: @neerajmasai this will break user's sites & starter templates as well, because till now targeted the immediate child of group, cover, column block but with this change the spacing will apply to all blocks which will increase inner spacing of all nested container which looks bad on the frontend. 

Image scenario of Cover block having column block inside it, so with this change the padding (let's say 4em) will apply to cover as well and to the columns as well which will become 8em of total spacings which breaks the UI completely. 

For the custom layout support try something from addon side instead of manipulating this stable logic.

neerajmasai: @imnavanath Okay, will implement accordingly. Thanks for pointing this out.

PR #5068, opened 2023-03-27
AST-3096 - Flyout menu issue with RTL version

yssambare12:
**Test Data: (If any)**  

Flyout menu issue with RTL version, when we enable rtl, then there is some clickable div is showing on corner.

**Steps To Reproduce:** 

For better understanding here I have created a short screencast for you to show you exactly how it works. [[Refer to this screencast](https://share.getcloudapp.com/NQu40EWg)]
1. Just enable RTL and click on toggle menu , you will see the issue in mobile devices .
2. There is clickable div is present for rtl version.

Any Other details:
### Checklist:
- [x] My code is tested


inc/builder/type/header/off-canvas/dynamic-css/dynamic.css.php
```diff
@@ -252,7 +252,7 @@ function astra_off_canvas_static_css() {
 		}
 		.ast-mobile-popup-drawer .ast-mobile-popup-inner {
 			width: 100%;
-			transform: translateX(100%);
+			transform: translateX(115%);
```
imnavanath, line 255: @yssambare12 why 115% only? Any specific reason?

yssambare12: Yes, @imnavanath this is add because when we set site to RTL mode that time some clickable element showing on left corner of screen, so for that we need to set this -115%.
https://d.pr/v/K3r9YC

PR #5082, opened 2023-04-04
AST-3122 - Fix: Slide Cart Goes Blank when Elementor Pro is enabled.

yssambare12:
**Test Data: (If any)**  

Slide Cart Goes Blank when Elementor Pro is enabled.

**Steps To Reproduce:** 

1. Install Elementor & Astra & woocommerse.
2. then set the cart to slide in .
3. now check the cart by adding some product inside this.

### Checklist:
- [x] My code is tested


inc/compatibility/woocommerce/class-astra-woocommerce.php
```diff
@@ -3339,6 +3339,13 @@ public function cart_link_fragment( $fragments ) {
 			$this->astra_get_cart_link();
 			$fragments['a.cart-container'] = ob_get_clean();
 
+			ob_start();
+
+			woocommerce_mini_cart();
+
+			$mini_cart = ob_get_clean();
+
+			$fragments['div.widget_shopping_cart_content'] = '<div class="widget_shopping_cart_content">' . $mini_cart . '</div>';
```
imnavanath, line 3348: will this create duplicate content in cart? only by updating priority will it work if we don't do this 3342-48 lines code?

yssambare12: @imnavanath The concern is right but with this No duplicate content will create because here we use the same array key for this. and just by updating priority and removing those line it will not work.  
the root cause is in elementor there is a same filter which is removing this that's why we increase the priority and did this modification.

imnavanath: Increasing priority only does not fix the issue?? Just asking, to check if we can ignore fragments updations.

yssambare12: No increasing priority will not fix this issue . because If we not update $fragments array, the front-end functionality that relies on this array to update the page which will not working correctly.

imnavanath: Okay got it.

changelog.txt
```diff
@@ -5,6 +5,7 @@ v4.1.3 (Unreleased)
 - Fix: Topic tags div visible multiple times and goes into infinite loop in case of BBPress.
 - Fix: Offcanvas - Builder component popup overlaps on customizer controls section.
 - Fix: WooCommerce - Widget's dropdown arrow is moving to top corner.
+- Fix: WooCommerce - Slide Cart Goes Blank when Elementor Pro is enabled.
```
imnavanath, line 8:
```suggestion
- Fix: Header Builder - Cart - Slide cart goes blank when Elementor Pro is active with latest version.
```

imnavanath (approved):

imnavanath:
PR for this issue - https://github.com/brainstormforce/astra/issues/5083

PR #5086, opened 2023-04-05
AST-3120 - Fix: Missing form fields on BuddyPress registration page with Astra.

yssambare12:
### Description

Missing form fields on BuddyPress registration page with Astra.

### Screenshots
https://d.pr/i/WzMGqO

**Steps to replicate -**

1. Install Astra & BuddyPress
2. Go to setting and click on to anyone can regristar box.
3. Go to pages and check the buddpress registration page on frontend and in incognito mode.
4. Registration or login fields need to be shown but its not.

**Note** - While doing this changes for VIPCS TASK in excerpt related some modification happened that's why this issue comes in a front - https://d.pr/i/E1Xp5t

### Checklist:
- [x] My code is tested


inc/markup-extras.php
```diff
@@ -1373,7 +1373,7 @@ function astra_entry_header_class( $echo = true ) {
 		$thumb_markup     = astra_get_post_thumbnail( '', '', false );
 		$post_meta_markup = astra_get_post_meta( astra_get_option( 'ast-dynamic-single-' . $post_type . '-metadata', array( 'comments', 'author', 'date' ) ) );
 		$single_structure = 'page' === $post_type ? astra_get_option( 'ast-dynamic-single-page-structure', array( 'ast-dynamic-single-page-image', 'ast-dynamic-single-page-title' ) ) : astra_get_option( 'ast-dynamic-single-' . esc_attr( $post_type ) . '-structure', array( 'ast-dynamic-single-' . $post_type . '-title', 'ast-dynamic-single-' . $post_type . '-meta' ) );
-		$get_the_excerpt  = get_the_excerpt();
+		$get_the_excerpt  = is_buddypress() ? '' : get_the_excerpt(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
imnavanath, line 1376: This will give fatal error for calling undefined function for users who don't have BBPress on their add extra check before that. 
BTW what `is_buddypress` is returning? and how this will fix this issue?

yssambare12: @imnavanath Sorry My bad, I was trying to check if buddy press is active or not - code is updated accordingly now its the code will not throwing a fatal error.  if someone is not using the Buddypress plugin and the bp_is_active() function is not defined,

In this case, the function_exists('bp_is_active') check will return false, which means that the second option after the : symbol will be executing directly.

yssambare12: Ok Looking into this issue - debugging more.

yssambare12: @imnavanath PR Updated please check now.

imnavanath (approved):

PR #5163, opened 2023-05-26
AST-3203 - Fix: Sticky Header feature is overwritten by the Checkout dropdown fields

yssambare12:
**Test Data: (If any)**  

Sticky Header feature is overwritten by the Checkout dropdown fields. 

**Steps To Reproduce:** 

1. Enable the Sticky Header (Doesn’t matter if it’s the Primary, Above, or Below header).
2. Go to WooCommerce Checkout
3. Click in a dropdown field (Country or State for example).
4. Scroll down the page.

**Screencast: [CleanShot 2023-05-23 at 4.05.24](https://share.cleanshot.com/GJ6vDrKG)**

### Checklist:
- [x] My code is tested


nathanrodrigues2111 (approved):

inc/compatibility/woocommerce/class-astra-woocommerce.php
```diff
@@ -2405,6 +2405,16 @@ public function add_styles() {
 				);
 			}
 
+			if ( is_checkout() && ! is_wc_endpoint_url( 'order-received' ) ) {
+				$css_output .= astra_parse_css(
+					array(
+						'.select2-dropdown' => array(
```
imnavanath, line 2411: will this cause any issue on form fields? if any extrenal plugin adds select2 fields then also it can break. If this fix is correct so can we make this very specific to that particular element where the issue there?

imnavanath: as this global selector can break other select2 z-index as well

yssambare12: @imnavanath I removed the previous implementation as I found a more appropriate place to add this property. With the new implementation, this will be more specific which will not be causing any other issues.

assets/css/unminified/compatibility/woocommerce/woocommerce-rtl.css
```diff
@@ -3211,6 +3211,7 @@ p.demo_store[data-position="bottom"],
 .woocommerce .select2-container .select2-dropdown,
 .woocommerce-page .select2-container .select2-dropdown {
   border: 1px solid var(--ast-border-color);
+  z-index: 98;
```
imnavanath, line 3214: Will this cause any issue on other select2 fields, like on checkout page, my account page etc...??

imnavanath (approved):

PR #5164, opened 2023-05-30
AST-3211 - Fix: Fatal error: Uncaught TypeError: Unsupported operand types: On frontend and customizer.

yssambare12:
**Test Data: (If any)**  

We are getting some Fatal error: Uncaught TypeError: Unsupported operand types: On frontend and customizer .

**Screenshot** - https://d.pr/i/IURQeh 

**Steps To Reproduce:** 

Just import this customizer JSON file and you will noticed the error.

### Checklist:
- [x] My code is tested

### DIFF
--- a/inc/class-astra-dynamic-css.php
+++ b/inc/class-astra-dynamic-css.php
@@ -3629,7 +3629,7 @@ if ( ! class_exists( 'Astra_Dynamic_CSS' ) ) {
 			 *
 			 *   - Sidebar Positions CSS
 			 */
-			$secondary_width = astra_get_option( 'site-sidebar-width' );
+			$secondary_width = absint( astra_get_option( 'site-sidebar-width' ) );
 			$primary_width   = absint( 100 - $secondary_width );
 			$meta_style      = '';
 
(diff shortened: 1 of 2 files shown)


inc/class-astra-dynamic-css.php
```diff
@@ -3629,9 +3629,9 @@ public static function return_meta_output( $dynamic_css, $dynamic_css_filtered =
 			 *
 			 *   - Sidebar Positions CSS
 			 */
-			$secondary_width = astra_get_option( 'site-sidebar-width' );
-			$primary_width   = absint( 100 - $secondary_width );
-			$meta_style      = '';
+			$secondary_width = absint( astra_get_option( 'site-sidebar-width' ) );
```
nathanrodrigues2111, line 3632: Any reason we are using absint for primary_width but strval for secondary_width @yssambare12

yssambare12: @nathanrodrigues2111 To resolve the main fatal error we need to do this changes as  value is properly retrieved and converted to an integer before calculating the $primary_width.

and the second is use strval because I get this code analysis error - [](https://d.pr/i/kFLB9v)
to resolve this need to I add this strval as the function expects a string as the first argument, but an integer was provided instead so we need to make sure that the $secondary_width variable is a string before passing it to astra_get_css_value().

nathanrodrigues2111: okay @yssambare12

nathanrodrigues2111 (approved):

PR #5166, opened 2023-05-30
 AST-3204 - difference-in-font-size-for-the-mobile-menu-between-the-front-end-and-back-end

racingfoxs:
### Description
When setting font size from Customize>>Header Builder>>Off-Canvas Menu>>Design there’s a difference in sizes displayed in the front-end and back-end.
Basically rem it applied if we select px on backend setting.
And if we choose em it is working fine.

### Screenshots

Backend snapshot - https://d.pr/i/pXWxen 
Frontend snapshot - https://d.pr/i/2Dfg5H 
Screen Recording - https://d.pr/v/Rolbvb
Ticket URL - https://secure.helpscout.net/conversation/2235719209/386337

### Types of changes

Bug Fix

### How has this been tested?
Tested with px and em on off canvas menu setting.
Tested with changing the body typography. 

### Checklist:
- [x] My code is tested


yssambare12 (approved):

inc/core/common-functions.php
```diff
@@ -177,7 +177,7 @@ function astra_get_font_css_value( $value, $unit = 'px', $device = 'desktop' ) {
 					$value            = intval( $value );
 					$fonts            = array();
 					$body_font_size   = astra_get_option( 'font-size-body' );
-					$fonts['desktop'] = ( isset( $body_font_size['desktop'] ) && '' != $body_font_size['desktop'] ) ? $body_font_size['desktop'] : 15;
+					$fonts['desktop'] = ( isset( $body_font_size['desktop'] ) && '' != $body_font_size['desktop'] ) ? $body_font_size['desktop'] : 16;
```
imnavanath, line 180: This function call every time for each font-size, as we are updating here fallback this can cause problem in some case, which can be problematic. Any other way to tackle this?? or else if not so major and just a preview visual change then it can be skip.

yssambare12: @imnavanath We can skip this for now as this is not a major issue we can say. If we got multiple queries regarding this then we can reopen for this.

PR #5188, opened 2023-06-23
AST-3269 - Fix: Removed extra Read time option from theme which visible even after addon is disable.

yssambare12:
### Description

Removed extra Read time option from theme which visible even after addon is disable.

**Steps To Reproduce:** 

1. Keep only the Astra theme active
2. Navigate to Customize > Blog > Blog/Archive and enable the read time.
3. It does not have any effect.
4. If Astra Pro is activated with Blog Pro, then it has an effect and correctly appears.

### Screenshots

- https://d.pr/v/9pNVsk

### Checklist:
- [x] My code is tested

### DIFF
--- a/.github/workflows/pull-request-title.yml
+++ b/.github/workflows/pull-request-title.yml
@@ -16,7 +16,7 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


yssambare12:
@imnavanath Do you think we needed a changelog for this?

PR #5264, opened 2023-07-25
AST-3349 - Fix: Spectra Compatibility CSS which overrides Spectra Block Spacing Removed. 

neerajmasai:
### Description

- Earlier, during our Spectra Compatibility Release, in order to match up Frontend and Editor, we had written CSS Specific to Spectra blocks to align with Editor Page Title, but it is now causing issues with the Block. As Block Specific CSS is not overriding the Theme CSS. Therefore, we need to remove the CSS.
- JIRA : https://brainstormforce.atlassian.net/browse/AST-3349

### Screenshots

- [Screen Capture on 2023-07-25 at 16-29-06.mp4](https://d.pr/v/Pmb6n0)

### Types of changes

- Bug Fixes :- Spectra Block Specific Padding Left Not Overidding as Expected.
- Breaking Change :- Alignment with the Editor Page Title.

### How has this been tested?

- Install Spectra.
- Add a new Post/Page.
- Spectra Block Specific Padding Left should apply correctly as expected in frontend.
- Refer to video :- - [Screen Capture on 2023-07-25 at 16-29-06.mp4](https://d.pr/v/Pmb6n0)

### Checklist:
- [x] My code is tested


inc/dynamic-css/block-editor-compatibility.php
```diff
@@ -477,18 +477,9 @@ function astra_load_modern_block_editor_ui( $dynamic_css ) {
 	// Spectra Compatibility - Container block alignment with page title for container layouts.
 	if ( $spectra_gutenberg_compat_css ) {
 		$dynamic_css .= '
-			.ast-separate-container .entry-content .uagb-is-root-container {
-				padding-' . esc_attr( $ltr_left ) . ': 0;
-			}
```
yssambare12, line 482: @neerajmasai we are removing this so this is not causing an issue on the editor preview because I remember we add this padding and left 0 for list item adjustment.

neerajmasai: @yssambare12 In any case, currently it is overriding the spectra block spacing setting so that is an issue, we need to find alternative to fix the other cases without using spectra selectors. We can have a call to discuss the same.

PR #5326, opened 2023-08-15
AST-3404 - Change Header logo color.

nathanrodrigues2111:
### Description
Jira : https://brainstormforce.atlassian.net/browse/AST-3404

### Types of changes

### Checklist:
- [x] My code is tested

### DIFF
--- a/inc/markup-extras.php
+++ b/inc/markup-extras.php
@@ -205,6 +205,22 @@ if ( ! function_exists( 'astra_logo' ) ) {
 				add_filter( 'wp_get_attachment_image_src', 'astra_replace_header_logo', 10, 4 );
 			}
 
+			$header_logo_color = astra_get_option( 'header-logo-color' );
+
+			if ( $header_logo_color ) {
+				
+				if ( 0 === strpos( $header_logo_color, 'var(--' ) ) {
+					$agp       = new Astra_Global_Palette();
+					$svg_color = astra_hex_to_rgb( $agp->get_color_by_palette_variable( $header_logo_color ) );
+				} elseif ( ! preg_match( '^(#)((?:[A-Fa-f0-9]{3}){1,2})$', $header_logo_color ) ) {
+					$svg_color = astra_hex_to_rgb( $header_logo_color );
+				} else {
+					$svg_color = astra_split_rgba( $header_logo_color );
+				}
+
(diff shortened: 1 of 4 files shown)


inc/markup-extras.php
```diff
@@ -1705,3 +1721,122 @@ function astra_bbpress_issue( $value ) {
...
+
+/**
+ * Converts HEX to RGB.
+ *
+ * @since x.x.x
+ *
+ * @param string $hex Hex color.
+ * @return array split version of rgb.
+ */
+function astra_hex_to_rgb( $hex ) {
```
imnavanath, line 1805: We already have [astra_hex_to_rgba](https://github.com/brainstormforce/astra/blob/293c9cb534fa56cb305e0c94e154164d149e069b/inc/core/common-functions.php#L1105-L1157) function can we utilize it, instead of creating new one. It also has optional $opacity parameter with marking as false it will return `rgb()` only.

nathanrodrigues2111: The return on both the function is different the one you suggested return rgb( ... ) or rgba( ... )

But what i need is array( 'r' -> '', 'g' => '', 'b' =>  '' )

So that is why i created a new function for it  @imnavanath

imnavanath: Okay got it. 👍 
Let's move this 2 new functions in same common-functions.php file.

nathanrodrigues2111: okay @imnavanath

nathanrodrigues2111: Moved all common functions @imnavanath

inc/markup-extras.php
```diff
@@ -1705,3 +1705,27 @@ function astra_bbpress_issue( $value ) {
...
+		astra_render_svg_mask( 'ast-img-color-filter', 'header_logo_svg_color', $header_logo_color );
+	}
+
+	if ( $transparent_header_logo_color ) {
+		/** @psalm-suppress UndefinedFunction  */ // phpcs:ignore Generic.Commenting.DocComment.MissingShort
+		astra_render_svg_mask( 'ast-img-color-filter-2', 'header_logo_svg_color', $transparent_header_logo_color );
+	}
+}
+
+add_filter( 'wp_footer', 'render_header_svg_mask' );
```
imnavanath, line 1731:
```suggestion
add_action( 'wp_footer', 'render_header_svg_mask' );
```

imnavanath (approved):

PR #5341, opened 2023-08-23
AST-3432 - Fix: Issue with WooCommerce Single Product Layout not applying correctly.

neerajmasai:
### Description

- WooCommerce Single Product page layout not applying correctly as expected when all WooCommerce Settings set to default.
- Also added Content Style and Sidebar Style support in few places where it is supposed to be added.
- JIRA - https://brainstormforce.atlassian.net/browse/AST-3432

### Screenshots

https://d.pr/v/DbHEHI

### Types of changes

Bug Fix

### How has this been tested?

- Check the Default Cases for Single Product.
- Check the Meta Cases for Single Product.
- Check other layouts for Single Product.
- Check the migration cases for Single Product.
- Check other pages like WooCommerce Shop etc to be sure.
- Check Content Color and Site Color settings along with backward cases.

### Checklist:
- [x] My code is tested


inc/extras.php
```diff
@@ -227,7 +227,7 @@ function astra_toggle_layout( $new_content_option, $level, $post_id = false, $ol
 			if ( 'meta' === $level && ! $migrated_user && 'set' !== $meta_key && $old_meta ) {
 				$current_layout = $old_meta;
 			} else {
-				$current_layout = '';
+				$current_layout = $dynamic_layout_option;
```
imnavanath, line 230: Will this cause any problem for other working layouts?

neerajmasai: It was causing a problem earlier:

For normal pages:
It should be ''

For WooCommerce pages:
it should be 'default'

Instead of setting explicitly to '' or 'default' we are setting it to `$dynamic_layout_option`

Why?
Because for default case in switch statement `$dynamic_layout_option` will automatically take the correct value i.e '' or 'default' depending on the page whether woocommerce or normal (posts, blogs etc)

imnavanath: Let's get this tested.

imnavanath (approved):

inc/extras.php
neerajmasai: @imnavanath Asked QA to retest the WooCommerce module and related config as well for more clarity

nathanrodrigues2111 (approved):

PR #5344, opened 2023-08-24
AST-3249 - Feat: Added new featured image options for single dynamic layouts

imnavanath:
### Description
- New featured image options for 
- 1. Image Ratio
- 2. Hover Effect
- 3. Bordered Image
- 4. Image Size

### Screenshots
- https://d.pr/v/KPWzhm

### Types of changes
- New feature

### How has this been tested?
- New options for featured image

### Checklist:
- [x] My code is tested


assets/js/unminified/customizer-controls.js
```diff
@@ -75,6 +75,46 @@
...
+		 * Toggle sub control visibility CSS.
+		 *
+		 * @since x.x.x
+		 */
+		subControlsToggleCSS( controlValue, dependents ) {
+			$.each( dependents, function( controlOption, dependentSubControls ) {
+				$.each( dependentSubControls, function( dependentIndex, subControl ) {
+					// Remove old.
+					jQuery( 'style#ast-sub-control-' + subControl ).remove();
+					// Add new.
```
yssambare12, line 90: @imnavanath If I'm not wrong currently we are removing old styling only when new styling is applied, but if the user tries this option 2 or 3 times then with every interaction it will continuously add the styling and these are unused styles.
So in my opinion can we remove old styles outside of this loop?

inc/customizer/extend-custom-controls/src/ast-number/ast-number-component.js
```diff
@@ -0,0 +1,32 @@
+import PropTypes from 'prop-types';
+import { useState } from 'react';
+
+const AstNumberComponent = props => {
+	const getValue = ( '' === props.control.setting.get() ) ? '' : props.control.setting.get();
+	const [ numberState, setNumberState ] = useState( getValue );
+	let htmlLabel = null;
```
yssambare12, line 7: @imnavanath Curious to know why we initially set the label null and then conditionally rendering this.

imnavanath: Yes because sometimes we don't require title/label for control. Real case - https://d.pr/i/OhIK5S -> https://d.pr/i/q3mtXT That's why.

assets/js/unminified/customizer-controls.js
imnavanath: In trigger event only we get all the sub-controls which have context attached. If we want to remove this from the outside loop, then in that case every time we need to keep records of the contextual sub-controls list. Every time in every case (like dynamic customizer controls) it won't be possible & feasible.
That's why

PR #5424, opened 2023-09-27
AST-3422 - Feature Secondary Button Outline Styling Options.

neerajmasai:
### Description

User Story:

- As a PO/End-user
- I want to style my site’s outline buttons
- Add design styling options which will be working for outline buttons
- Make this functionality extendable 
- We will cover wp outline buttons by core but this styling should also work on components/elements where “ast-outline-button“ class will be use.
- In this way we provide users extendability to have this option usage in their way & custom buttons.
- So that  it will be differentiating from solid buttons, and the side-by-side buttons looks elegant.

- JIRA - https://brainstormforce.atlassian.net/jira/software/projects/AST/boards/10?assignee=62c6a337ce5a604dbfb504b8&selectedIssue=AST-3422

### Screenshots

https://d.pr/i/fgH1ZJ

### Types of changes

New Feature

### How has this been tested?

- Two new tabs should be visible - Primary, Secondary in Global > Button.
- Create a new Button with its WordPress Styling set to “Outline”.
- For this button, all the new Secondary Options should work as expected.
- Note: If Secondary Options are set to empty or blank, then some of the values will be inherited from Primary section (because earlier as we had a common setting for all the buttons: Solids + Outline)
If we set any Secondary Options they should apply as expected in Customizer, Frontend and the Editor.
- Older Users Compatibility :- Switch to previous version of Astra (master), and set up some options for Buttons and now switch to the latest version, here you will observe couple of things:
             a. Buttons should remain as it in frontend (no button styling should break for Solids as well as Outline)
             b. Button presets will be reset in customizer will be reset, as now Button Presets are divided in two parts Primary and Secondary, so they will apply as per whatever value we select and will only work for that type, either Primary or Secondary.
- Create a new Custom HTML Block - Add a button and a class “ast-outline-button”. Secondary option values should also apply to this button.
- Test by installing some Starter templates on older version, switch to new version (latest) and ensure that no button styles are breaking.
- Test with Surecart as well.
- Test Responsive options.

### Checklist:
- [x] My code is tested


changelog.txt
```diff
@@ -1,5 +1,6 @@
 v4.3.2 (Unreleased)
 - New: Introducing a feature of Sticky Sidebar functionality for enhanced user experience. (Needs Doc)
+- New: Customizer - Secondary (Outline) Button Styles. (Needs Doc)
```
imnavanath, line 3:
```suggestion
- New: Introducing Primary & Secondary Button styles under Global - Buttons. (Needs Doc)
```

inc/class-astra-dynamic-css.php
```diff
@@ -1776,6 +1776,145 @@ public static function return_output( $dynamic_css, $dynamic_css_filtered = '' )
...
+			);
+
+			/* Parse CSS from array() -> Desktop */
+			$parse_css .= astra_parse_css( $outline_button_css_desktop );
+
+			/* Parse CSS from array() -> Tablet */
+			$parse_css .= astra_parse_css( $outline_button_css_tablet, '', astra_get_tablet_breakpoint() );
+
+			/* Parse CSS from array() -> Mobile */
+			$parse_css .= astra_parse_css( $outline_button_css_mobile, '', astra_get_mobile_breakpoint() );
```
imnavanath, line 1917: Can we make this entire change in new dynamic CSS file as this file is becoming very huge.

inc/customizer/configurations/buttons/class-astra-customizer-button-configs.php
```diff
@@ -35,199 +35,428 @@ class Astra_Customizer_Button_Configs extends Astra_Customizer_Config_Base {
...
-								'top'    => 0,
-								'right'  => 0,
-								'bottom' => 0,
-								'left'   => 0,
-							),
-							'button-radius-fields' => array(
-								'desktop'      => array(
+			
+			$_configs = array();
+			for ( $index=0, $id=''; $index < 2; $index++, $id = 'secondary-' ) {
```
imnavanath, line 40: Updating $id outside `for` line will be better for sake of minor PHP versions. You can make it conditional inside loop.

neerajmasai: Sure

inc/customizer/configurations/buttons/class-astra-customizer-button-configs.php
```diff
@@ -838,9 +871,10 @@ public function register_configuration( $configurations, $wp_customize ) {
 					),
 				);
 				$_configs = array_merge( $_configs, $_trans_config );
-
+
```
imnavanath, line 874:
```suggestion
```

imnavanath (approved):

PR #5533, opened 2023-11-13
AST-3616 - Embed(Twitter, soundcloud, etc) block not working on page.

Sudhanshu-1508:
### Description

[Task](https://brainstormforce.atlassian.net/browse/AST-3616)

### Types of changes

Non-breaking changes
### Checklist:
- [x] My code is tested
- [x] My code passes the PHPCS tests
- [x] My code follows accessibility standards 
- [x] My code has proper inline documentation 
- [x] I've included any necessary tests 
- [x] I've included developer documentation 
- [ ] I've added proper labels to this pull request

PR #5541, opened 2023-11-15
AST-3616 - Fix: astra_responsive_oembed_wrapper_enable filter

IngoWinter:
the first parameter for the `responsive_oembed_wrapper` method should be `$matches[1]` but it is `''`. That renders the `astra_responsive_oembed_wrapper_enable` Filter in [/inc/class-astra-after-setup-theme.php](https://github.com/brainstormforce/astra/blob/v4.4.0/inc/class-astra-after-setup-theme.php#L247) useless.


imnavanath:
Considered this fix with updated code - https://github.com/brainstormforce/astra/pull/5533 Thanks @IngoWinter for your contribution.

PR #5635, opened 2023-12-21
AST-3696 - Smooth Scroll to ID Causes an Anchor Link goes to The Wrong Anchor Link 

Sudhanshu-1508:
### Description

[Task](https://brainstormforce.atlassian.net/browse/AST-3696)
### Checklist:
- [x] My code is tested


assets/js/unminified/navigation.js
```diff
@@ -661,7 +661,11 @@ astScrollToTopHandler = function ( masthead, astScrollTop ) {
 
 			for (const link of links) {
 
-				if (link.hash !== "") {
+				
+				if (link.href.split('#')[0] !== location.href.split('#')[0]) {
+					// Store the hash
+					sessionStorage.setItem('scrollTo', link.hash);
```
garvits-bsf, line 667: Do we need `sessionStorage` here? In this context, we can assign `link.hash` value to a variable which will be more memory efficient than sessionStorage.

assets/js/unminified/navigation.js
```diff
@@ -696,6 +700,34 @@ astScrollToTopHandler = function ( masthead, astScrollTop ) {
 		}
 	}
 
+	window.addEventListener('DOMContentLoaded', (event) => {
```
garvits-bsf, line 703: If this block of code is only required when `astra.is_scroll_to_id` is `true` then let's keep this block of code inside the `if ( astra.is_scroll_to_id ) ` condition, so that we do not initialize this particular block of code otherwise.

garvits-bsf (approved):

assets/js/unminified/navigation.js
```diff
@@ -655,17 +655,22 @@ astScrollToTopHandler = function ( masthead, astScrollTop ) {
...
+		let linkHash = '';
 		const links = document.querySelectorAll('a[href*="#"]:not([href="#"]):not([href="#0"]):not([href*="uagb-tab"]):not(.uagb-toc-link__trigger):not(.skip-link):not(.nav-links a):not([href*="tab-"])');
 		if (links) {
 
 			for (const link of links) {
 
-				if (link.hash !== "") {
+				if (link.href.split('#')[0] !== location.href.split('#')[0]) {
+					// Store the hash
+					linkHash = link.hash;
```
imnavanath, line 667: If we have multiple `link.hash` then at every instance `linkHash` will get override and it will take last instance only. Have we checked the case?

imnavanath (approved):

imnavanath:
Added here - https://github.com/brainstormforce/astra/pull/5677

